PR #9, opened 2024-12-02
Implement functionality in the function yOCTFindTissueSurface

alejoruge:


Processing/yOCTFindTissueSurface.m
```diff
@@ -27,9 +28,91 @@
...
+
+
+%% Generate Heatmap of Surface Height for user visualization
+if isVisualize
+    figure;
+    imagesc(x, y, surfacePosition);
+    colormap(flipud(jet));
+    colorbar;
+    set(gca, 'YDir', 'normal');
+    title(['Surface position in ', dim.x.units, ' (View from the top)']);
```
MyYo, line 115: Surface position should be in dim.z.units (which may or may not be the same as dim.x.units)

Processing/yOCTFindTissueSurface.m
```diff
@@ -27,9 +28,91 @@
...
+    xlabel(['X-axis (', dim.x.units, ')']); 
+    ylabel(['Y-axis (', dim.x.units, ')']);
```
MyYo, line 117: change to dim.y.units

MyYo (changes requested):
Thanks, looks good but requires two small changes.
Does the tester passes? It's recommended to commit code only once tests pass.

alejoruge:
Hi, thank you!

Yes, it passes now! I did this:

I modified the tester only so that it simulates a tissue surface (instead of just a single line, it has some data), so that the function can work with this simulated data.

Here's a picture of what one simulated slice would look like with this change for you to take a look: 
SurfacePosition_Tester


Now in regards of the function I made this change:

Thanks to the tester, I realized I needed to replace my manual micron-per-pixel calculation with directly using the indexed Z values for surfacePosition in units, ensuring the distances reflect the actual measurements. Thank you so much!

Processing/test_yOCTFindTissueSurface.m
```diff
@@ -14,8 +14,10 @@ function createDummyDataset(testCase)
             % Create a dummy dataset with a scattering interface
             testCase.simulatedSurfacePositionZ_pix = 500;
             speckleField = zeros(1024,100,200)+10; %z,x,y
-            speckleField(testCase.simulatedSurfacePositionZ_pix,:,:) = 1000;
-    
+
+            speckleField(testCase.simulatedSurfacePositionZ_pix:end, :, :) = ...
+                10 + 990 * abs(randn(1024 - testCase.simulatedSurfacePositionZ_pix + 1, 100, 200));
```
MyYo, line 19: Please avoid using "randn" - random numbers are not good for testing because if something fails due to random value it's hard to reproduce.

Processing/yOCTFindTissueSurface.m
```diff
@@ -2,19 +2,20 @@
...
+%   logMeanAbs, dimensions - data structures generated by yOCTProcessTiledScan.
 %       Converted to log scale
 %   isVisualize - set to true to generate image heatmap visualization
-%       figure. Default is false
+%       figure. Default is false.
 % OUTPUTS:
 %   - surfacePosition- 2D matrix. dimensions are (y,x). What
 %       height is image surface (measured from the top of the OCT scan).
-%       physical dimensions of surfacePosition are the same as dimensions.z.units
+%       Physical dimensions of surfacePosition are the same as dimensions.z.values
```
MyYo, line 14: Can you explain the sentence?

I recommend changing to:
% Physical **units** of surfacePosition are the same as dimensions.z.**units**

Processing/yOCTFindTissueSurface.m
```diff
@@ -2,19 +2,20 @@
...
 %   - x,y are the x,y positions that corresponds to surfacePosition(y,x).
 %       Units are the same as dimensions.x.units
 
+
```
MyYo, line 18: Extra line, can be removed.

MyYo (changes requested):
Thanks. A few small remaining comments

Processing/test_yOCTFindTissueSurface.m
MyYo: If you think random numbers reproduce the results better use rng(1); before generating the random numbers.
This will make sure that random numbers start from the same seed (1 in this case)

alejoruge:
I apologize for the random numbers. To get a more stable, consistent simulation, I can overwrite one reference line (after yOCTInterfToScanCpx has processed the data) and replicate it multiple times (20 in this example), ensuring no random fluctuation.

Here’s an example (I committed it like this):

passing



As an alternative, I can also add a seed if you prefer... Please let me know what you think..

Processing/test_yOCTFindTissueSurface.m
```diff
@@ -14,12 +14,14 @@ function createDummyDataset(testCase)
...
             speckleField = zeros(1024,100,200)+10; %z,x,y
-            speckleField(testCase.simulatedSurfacePositionZ_pix,:,:) = 1000;
-    
+            speckleField(testCase.simulatedSurfacePositionZ_pix, :, :) = 1000;
+        
             [interf, dim] = yOCTSimulateInterferogram(speckleField);
             [cpx, dim] = yOCTInterfToScanCpx(interf, dim);
             logMeanAbs_tmp = log(abs(cpx));
             testCase.logMeanAbs = logMeanAbs_tmp;
+            refSliceValue = testCase.logMeanAbs(testCase.simulatedSurfacePositionZ_pix, :, :);
```
MyYo, line 23: Move this to row 17. logMeansAbs shouldn't be changed after simulating the speckle field.

MyYo (changes requested):
It's okay to leave as it was just add "rng(1);" in line 17 to make sure that random numbers have the same value every time you execute the function.

Does it make sense why I want to use rng(1)?

alejoruge:
Yes, thank you. I added the seed now on line 17

MyYo (approved):
Thanks Looks good to me!

PR #10, opened 2024-12-12
Use precise zDepth measurement using user-defined tissue interface

alejoruge:
Enhance z dimensions handling for focus positioning in reconstructed images and use a dedicated zero-depth index for easier referencing later


MyYo (approved):
Looks perfect. Thank you so much!

PR #17, opened 2025-02-26
Minor Improvements (see notes for details)

aidanvanvleck:
Demo_CheckFocusInTileScan:
1) Simplified input folder
2) Save output file with sample name and input parameters 3) Save output file to the input folder

Demo_DispersionCorrectionManual:
1) Removed old default filepath

Demo_PhotobleachSquare:
1) Remove default bias that caused shifting of photobleach square off-center 2) Added option to toggle L-shape
3) Added option to toggle crosshair pattern in center of square

Demo_Process_3D:
1) Removed old default file path
2) Updated default dispersionQuadraticTerm

Demo_ScanAndProcess_3D:
1) Standardized default dispersionQuadraticTerms
2) Added 40x, OCTP900 dispersionQuadraticTerm to default list

### DIFF
--- a/Demo_CheckFocusInTileScan.m
+++ b/Demo_CheckFocusInTileScan.m
@@ -5,20 +5,30 @@
 %% Inputs
 
 % Input folder
-tiledScanInputFolder = './'; % Make sure folder path ends with "/" to signal this is a folder
+sampleFolder = ''; % Input your main sample folder here. This folder should contain a folder called 'OCTVolume' with your OCT data. This is where your output file will be saved.
+tiledScanInputFolder = fullfile(sampleFolder, 'OCTVolume', filesep); % filesep should automatically set the appropriate file separator
 
 % Processing parameters
-dispersionQuadraticTerm=-2.059e8;
-focusSigma = 20; % When stitching along Z axis (multiple focus points), what is the size of each focus in z [pixels]. OBJECTIVE_DEPENDENT: for 10x use 20, for 40x use 20 or 1
+dispersionQuadraticTerm=-1.482e8;
+focusSigma = 6; % When stitching along Z axis (multiple focus points), what is the size of each focus in z [pixels]. OBJECTIVE_DEPENDENT: for 10x use 20, for 40x use 20 or 1
 applyPathLengthCorrection = true;
 
 % For all B-Scans, this parameter defines the depth (Z, pixels) that the focus is located at.
 % If set to NaN, yOCTFindFocusTilledScan will be executed to request user to select focus position.
 focusPositionInImageZpix = NaN;
 
-% Output
-numberOfZScansToOutput = 10; % Set to 1e5 to output all scans
-output_figure = 'out.tif';
+%% Output
+numberOfZScansToOutput = 2; % Set to 1e5 to output all scans
+
+% Extract only the last folder name from the sampleFolder path
+sampleName = regexp(sampleFolder, '[^\\/]+$', 'match', 'once');
+
+% Format the output figure name using the required inputs
+output_figure = sprintf('CheckFocusInTileScan__%s__dQ%.3e_fS%d_f%d.tif', ...
+    sampleName, dispersionQuadraticTerm, focusSigma, focusPositionInImageZpix);
+
+% Set the output path inside sampleFolder
+output_figure = fullfile(sampleFolder, output_figure);
 
 
 %% Preprocess
--- a/Demo_DispersionCorrectionManual.m
+++ b/Demo_DispersionCorrectionManual.m
@@ -5,8 +5,7 @@ close all;
 
 %% Inputs
 %Ganymede
-filePath = ['\\171.65.17.174\MATLAB_Share\Jenkins\myOCT Build\TestVectors\' ...
-    'Wasatch_3D'];
+filePath = [''];
 
 %% Pre-Process
 %Load Intef From file
--- a/Demo_Process_3D.m
+++ b/Demo_Process_3D.m
@@ -4,9 +4,8 @@
 
 %% Iputs
 %Wasatch
-filePath = ['\\171.65.17.174\MATLAB_Share\Jenkins\myOCT Build\TestVectors\' ...
-    'Wasatch_3D\'];
-dispersionQuadraticTerm = 100; %Use Demo_DispersionCorrection to find the term
+filePath = [''];
+dispersionQuadraticTerm = -1.514e08; %Use Demo_DispersionCorrection to find the term
 
 yFramesPerBatch = 1; %How many Y frames to load in a single batch, optimzie this parameter to save computational time
 
--- a/Demo_PhotobleachSquare.m
+++ b/Demo_PhotobleachSquare.m
@@ -17,15 +17,62 @@ octProbePath = yOCTGetProbeIniPath('40x','OCTP900'); % Select lens magnification
 % Pattern to photobleach. System will photobleach n lines from 
 % (x_start(i), y_start(i)) to (x_end(i), y_end(i))
 scale = 0.25; % Length of each side of the square in mm
-bias = 0.01; % Use small bias [mm] to make sure lines are not too close to the edge of FOV
-x_start_mm = [-1, +1, +1, -1, -1.2, -1.2]*scale/2+bias;
-y_start_mm = [-1, -1, +1, +1, +1.2, +1.2]*scale/2+bias;
-x_end_mm   = [+1, +1, -1, -1, -1.2, +0  ]*scale/2+bias;
-y_end_mm   = [-1, +1, +1, -1, -1.2, +1.2]*scale/2+bias;
+bias = 0.0; % Use small bias [mm] to make sure lines are not too close to the edge of FOV
+photobleach_L_shape = true; % Select "true" if you want to photobleach an L shape outside of the square. Select "false" if not.
+photobleach_crosshairs = false; % Select "true" if you want to photobleach a crosshair pattern through the center of the square. Select "false" if not.
 
 % Photobleach configurations
 exposure_mm_sec = 20; % mm/sec
 nPasses = 1; % Keep as low as possible. If galvo gets stuck, increase number
+
+
+%% Initialize empty arrays
+x_start_mm = [];
+y_start_mm = [];
+x_end_mm   = [];
+y_end_mm   = [];
+
+
+%% Define patterns
+% Square (Base)
+square_x_start = [-1, +1, +1, -1]*scale/2 + bias;
+square_y_start = [-1, -1, +1, +1]*scale/2 + bias;
+square_x_end   = [+1, +1, -1, -1]*scale/2 + bias;
+square_y_end   = [-1, +1, +1, -1]*scale/2 + bias;
+
+% L-shape extension
+L_x_start = [-1.2, -1.2]*scale/2 + bias;
+L_y_start = [+1.2, +1.2]*scale/2 + bias;
+L_x_end   = [-1.2, +0  ]*scale/2 + bias;
+L_y_end   = [-1.2, +1.2]*scale/2 + bias;
+
+% Crosshairs
+cross_x_start = [0, 1]*scale/2 + bias;
+cross_y_start = [1, 0]*scale/2 + bias;
+cross_x_end   = [0, -1]*scale/2 + bias;
+cross_y_end   = [-1, 0]*scale/2 + bias;
+
+
+%% Combine patterns dynamically
+if photobleach_L_shape
+    x_start_mm = [x_start_mm, square_x_start, L_x_start];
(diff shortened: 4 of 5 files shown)


Demo_CheckFocusInTileScan.m
```diff
@@ -5,20 +5,30 @@
...
 % Input folder
-tiledScanInputFolder = './'; % Make sure folder path ends with "/" to signal this is a folder
+sampleFolder = ''; % Input your main sample folder here. This folder should contain a folder called 'OCTVolume' with your OCT data. This is where your output file will be saved.
+tiledScanInputFolder = fullfile(sampleFolder, 'OCTVolume', filesep); % filesep should automatically set the appropriate file separator
 
 % Processing parameters
-dispersionQuadraticTerm=-2.059e8;
-focusSigma = 20; % When stitching along Z axis (multiple focus points), what is the size of each focus in z [pixels]. OBJECTIVE_DEPENDENT: for 10x use 20, for 40x use 20 or 1
+dispersionQuadraticTerm=-1.482e8;
+focusSigma = 6; % When stitching along Z axis (multiple focus points), what is the size of each focus in z [pixels]. OBJECTIVE_DEPENDENT: for 10x use 20, for 40x use 20 or 1
```
MyYo, line 13: The comment defines default value of 20 or 1 but not 6. what lens are you using 6 for?

Demo_CheckFocusInTileScan.m
```diff
@@ -5,20 +5,30 @@
 %% Inputs
 
 % Input folder
-tiledScanInputFolder = './'; % Make sure folder path ends with "/" to signal this is a folder
+sampleFolder = ''; % Input your main sample folder here. This folder should contain a folder called 'OCTVolume' with your OCT data. This is where your output file will be saved.
+tiledScanInputFolder = fullfile(sampleFolder, 'OCTVolume', filesep); % filesep should automatically set the appropriate file separator
```
MyYo, line 9: The implementation you provided is too rigid for the common use case. I can agree that 
tiledScanInputFolder = './' might be too lose so we can compromise on:

tiledScanInputFolder = './OCTVolume/'

Demo_CheckFocusInTileScan.m
```diff
@@ -5,20 +5,30 @@
...
 
 % For all B-Scans, this parameter defines the depth (Z, pixels) that the focus is located at.
 % If set to NaN, yOCTFindFocusTilledScan will be executed to request user to select focus position.
 focusPositionInImageZpix = NaN;
 
-% Output
-numberOfZScansToOutput = 10; % Set to 1e5 to output all scans
-output_figure = 'out.tif';
+%% Output
+numberOfZScansToOutput = 2; % Set to 1e5 to output all scans
```
MyYo, line 21: Whats the reason to change to 2?

Demo_CheckFocusInTileScan.m
```diff
@@ -5,20 +5,30 @@
...
+
+% Extract only the last folder name from the sampleFolder path
```
MyYo, line 23: Please keep the *%% input* section of the code clean of calculations.

Also please explain the rationale to not use 'out.tif'

Demo_DispersionCorrectionManual.m
```diff
@@ -5,8 +5,7 @@
 
 %% Inputs
 %Ganymede
-filePath = ['\\171.65.17.174\MATLAB_Share\Jenkins\myOCT Build\TestVectors\' ...
-    'Wasatch_3D'];
+filePath = [''];
```
MyYo, line 8: Use a filePath that works..

Demo_Process_3D.m
```diff
@@ -4,9 +4,8 @@
 
 %% Iputs
 %Wasatch
-filePath = ['\\171.65.17.174\MATLAB_Share\Jenkins\myOCT Build\TestVectors\' ...
-    'Wasatch_3D\'];
-dispersionQuadraticTerm = 100; %Use Demo_DispersionCorrection to find the term
+filePath = [''];
```
MyYo, line 7: Use file path that works

MyYo (changes requested):
Good stuff. Left some comments and a few issues to be fixed.

Demo_CheckFocusInTileScan.m
aidanvanvleck: I use focusSigma of 6 for CheckFocusInTileScan with the 40x specifically (not for actual reconstructions). This one isn't as important, its just what I always use and always end up changing before running.

Demo_Process_3D.m
aidanvanvleck: We don't have any file paths that work globally. We either run the code from a local drive or from Google Drive (via the Google Drive desktop app). 
The local drive files will be different from computer to computer. 
And Google Drive uses different drive letters (G:\ on my personal computer, but it varies from PC to PC)

I believe the current file path is going to a drive on the old server. I don't think it works currently, and we need to delete it every time we run it. So I figured just having it empty would be easiest (we currently have to delete the file path every time we want to run). 


This change is low priority though, so if you like it as-is, that's totally fine too

Demo_CheckFocusInTileScan.m
aidanvanvleck: Mmm fair enough - I was trying to standardize with the other scripts (Demo_ScanAndProcess_3D, Demo_Process_3D) which accept the entire sample folder. 

You're right though, if the sample folder has multiple OCTVolume folders it wouldn't work. I'll change it

Demo_CheckFocusInTileScan.m
aidanvanvleck: No reason, likely a typo. I will fix.

Demo_CheckFocusInTileScan.m
aidanvanvleck: Will move any calculations. 


A few reasons I don't like 'out.tif':
1. If you re-run the script multiple times (which we do a lot with this code since we try multiple input variations (focusSigma, focus position, dispersionQuadraticTerm, etc)), and you don't change the name of the previously generated file, it will be overwritten. It is very easy to overwrite a file causing confusion. 
2. Currently, there are a number of scripts that use these temporary file names (e.g. 'out.tif', 'tmp.tif'), and it is difficult to identify which is which at a glance. 


In the new system, the inputs used when generating the file will be saved automatically in the title. This helps with point 1 (as if you change any of the inputs, the output file name will change, and thus won't overwrite your previously generated file), and point 2 (as you can easily identify which script was run and which inputs/parameters were associated with each generated file). 

I do this manually right now, but that is prone to errors/typos/etc.

Demo_CheckFocusInTileScan.m
aidanvanvleck: We use 6 specifically here because it gives a pretty tight set of images around the focus. Since this script is used to help us determine the actual focus position, the tighter the generated XY image stack is around the focus, the higher our "resolution" is when verifying the actual focus position. 

I updated the comment to reflect this. If you disagree, totally fine, I will change it back.

Demo_CheckFocusInTileScan.m
aidanvanvleck: I added a new section "%% Output Calculations". I can move it down in the script if you'd like.

Demo_DispersionCorrectionManual.m
aidanvanvleck: Same as Demo_Process_3D, copying the response here:




We don't have any file paths that work globally. We either run the code from a local drive or from Google Drive (via the Google Drive desktop app).
The local drive files will be different from computer to computer.
And Google Drive uses different drive letters (G:\ on my personal computer, but it varies from PC to PC)

I believe the current file path is going to a drive on the old server. I don't think it works currently, and we need to delete it every time we run it. So I figured just having it empty would be easiest (we currently have to delete the file path every time we want to run).

This change is low priority though, so if you like it as-is, that's totally fine too

Demo_CheckFocusInTileScan.m
MyYo: okay. update code and commit again then

Demo_CheckFocusInTileScan.m
MyYo: Store file at the matlab folder

%% Output --> % Output
%% Output Calculations --> % Output file name and place under %% Preprocess

Demo_DispersionCorrectionManual.m
MyYo: filePath = '.\\';

Demo_Process_3D.m
MyYo: filePath = '.\\';

MyYo (changes requested):
Looks good

Demo_CheckFocusInTileScan.m
```diff
@@ -5,20 +5,31 @@
...
+%% Output Calculations
+% Extract only the last folder name from the sampleFolder path
+sampleName = regexp(sampleFolder, '[^\\/]+$', 'match', 'once');
+
+% Format the output figure name using the required inputs
+output_figure = sprintf('CheckFocusInTileScan__%s__dQ%.3e_fS%d_f%d.tif', ...
+    sampleName, dispersionQuadraticTerm, focusSigma, focusPositionInImageZpix);
+
+% Set the output path inside sampleFolder
+output_figure = fullfile(sampleFolder, output_figure);
```
MyYo, line 32: change to outputFigurePath

Matlab uses camel case

Demo_CheckFocusInTileScan.m
```diff
@@ -5,20 +5,31 @@
...
 numberOfZScansToOutput = 10; % Set to 1e5 to output all scans
-output_figure = 'out.tif';
+
+
+%% Output Calculations
+% Extract only the last folder name from the sampleFolder path
+sampleName = regexp(sampleFolder, '[^\\/]+$', 'match', 'once');
+
+% Format the output figure name using the required inputs
+output_figure = sprintf('CheckFocusInTileScan__%s__dQ%.3e_fS%d_f%d.tif', ...
```
MyYo, line 28: Change to otutputFigureFileName

MyYo (changes requested):
Thanks. A few last small edits

Demo_CheckFocusInTileScan.m
aidanvanvleck: Done

Demo_CheckFocusInTileScan.m
aidanvanvleck: Done

MyYo (changes requested):
Thanks, a few minor remaining issues

Demo_CheckFocusInTileScan.m
```diff
@@ -5,26 +5,37 @@
...
 applyPathLengthCorrection = true;
 
 % For all B-Scans, this parameter defines the depth (Z, pixels) that the focus is located at.
 % If set to NaN, yOCTFindFocusTilledScan will be executed to request user to select focus position.
 focusPositionInImageZpix = NaN;
 
 % Output
-numberOfZScansToOutput = 12; % Set to 1e5 to output all z stack planes
-output_figure = 'out.tif';
+numberOfZScansToOutput = 10; % Set to 1e5 to output all scans
```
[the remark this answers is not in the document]
aidanvanvleck: Done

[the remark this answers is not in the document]
MyYo: Value is still 10. Please change to 12

Demo_CheckFocusInTileScan.m
```diff
@@ -5,24 +5,34 @@
...
 
 % Input folder
-tiledScanInputFolder = './'; % Make sure folder path ends with "/" to signal this is a folder
+tiledScanInputFolder = './OCTVolume/' % Replace the . with your sample folder. Keep the /OCTVolume/ extension to 1) point to the correct folder within your sample folder and 2) to signal this is a folder. 
 
 % Processing parameters
-dispersionQuadraticTerm=-2.059e8;
-focusSigma = 10; % When stitching along Z axis (multiple focus points), what is the size of each focus in z [pixels]. OBJECTIVE_DEPENDENT: for 10x use 20, for 40x use 10 or 1
+dispersionQuadraticTerm=-1.482e8; % 40x, OCTP900 
+focusSigma = 12; % When stitching along Z axis (multiple focus points), what is the size of each focus in z [pixels]. OBJECTIVE_DEPENDENT: for 10x use 20, for 40x use 10 or 6.
```
MyYo, line 12: Revert back to 10 here..

MyYo (changes requested):

Demo_CheckFocusInTileScan.m
aidanvanvleck: Done

MyYo (approved):
Thanks looks good

PR #20, opened 2025-03-25
Enable BigTIFF saving to handle files >4GB

alejoruge:
Use 'w8' to save TIFF as BigTIFF and prevent size limit errors (>4GB)


MyYo (approved):
Looks good, thanks for doing that!

PR #21, opened 2025-04-01
Update Polynomial Coefficients for Probe - Olympus 10x.ini

alejoruge:
Calibrate the distortion correction coefficients for this specific lenses and probe.


MyYo (approved):
Thanks looks good!

PR #23, opened 2025-04-04
Calibrate 10x OCTP900 lenses

alejoruge:
Update Polynomial Coefficients for Probe - Olympus 10x - OCTP900.ini


MyYo (changes requested):
Are you sure this is for the OCTP900 scanning head and not the old scanning head?

alejoruge:
Yes. I used the 10x lenses on the OCTP900 probe to generate the coefficients and tested them on a fresh scan of a flat table (4 tiles stitched along the X axis) to confirm alignment. It looked like a straight line as opposed to the previous ones, see the result in the image attached:

10x_OCTP900_4_tiles

MyYo (approved):
Thanks! Looks good

PR #24, opened 2025-04-04
Fix z-depth validation & enforce zDepth=0 scan presence

alejoruge:
- Guarantee at least one scan at zDepth=0 for consistent referencing.
- Add error message if zDepths is missing/invalid in JSON/OCTVolume.
- Adjust z-alignment using scan at zDepth=0 via index search.

### DIFF
--- a/Processing/yOCTProcessTiledScan_createDimStructure.m
+++ b/Processing/yOCTProcessTiledScan_createDimStructure.m
@@ -46,7 +46,7 @@ elseif length(focusPositionInImageZpix) == 1
 else
     % One value for each depth
     if isempty(json.zDepths) || ~isnumeric(json.zDepths) % Validate presence of z-depths data
-        error('Missing or invalid zDepth scans found. Check JSON file or OCTVolume folder.');
+        error('ScanInfo.json reports zero scans (empty zDepths). This volume appears to be empty or failed.');
     end
     [~, idx] = min(abs(json.zDepths)); % Adjust z-dimensions using the focus index at zDepth 0
     dimOneTile.z.values = dimOneTile.z.values - dimOneTile.z.values(focusPositionInImageZpix(idx));
--- a/Processing/yOCTProcessTiledScan_createDimStructure.m
+++ b/Processing/yOCTProcessTiledScan_createDimStructure.m
@@ -48,8 +48,14 @@ else
     if isempty(json.zDepths) || ~isnumeric(json.zDepths) % Validate presence of z-depths data
         error('ScanInfo.json reports zero scans (empty zDepths). This volume appears to be empty or failed.');
     end
-    [~, idx] = min(abs(json.zDepths)); % Adjust z-dimensions using the focus index at zDepth 0
-    dimOneTile.z.values = dimOneTile.z.values - dimOneTile.z.values(focusPositionInImageZpix(idx));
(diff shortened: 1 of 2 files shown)


Processing/yOCTProcessTiledScan_createDimStructure.m
```diff
@@ -37,15 +37,19 @@
...
+    % No adjustment because no focus info is provided
 elseif length(focusPositionInImageZpix) == 1
-    % One value
+    % One single focus value; shift z-dimensions with this value
     dimOneTile.z.values = dimOneTile.z.values - dimOneTile.z.values(focusPositionInImageZpix);
 else
     % One value for each depth
-    dimOneTile.z.values = dimOneTile.z.values - dimOneTile.z.values(focusPositionInImageZpix(json.zDepths == 0));
+    if isempty(json.zDepths) || ~isnumeric(json.zDepths) % Validate presence of z-depths data
+        error('Missing or invalid zDepth scans found. Check JSON file or OCTVolume folder.');
```
MyYo, line 49: Error message is too generic.

MyYo (changes requested):
Looks good, small change requested.

Processing/yOCTProcessTiledScan_createDimStructure.m
alejoruge: I added this error just in case zDepths is empty — which really shouldn't happen unless no scans were made... I added it for safety, but let me know if you'd rather leave this error out..

Processing/yOCTProcessTiledScan_createDimStructure.m
```diff
@@ -37,15 +37,19 @@
...
-    % One value
+    % One single focus value; shift z-dimensions with this value
     dimOneTile.z.values = dimOneTile.z.values - dimOneTile.z.values(focusPositionInImageZpix);
 else
     % One value for each depth
-    dimOneTile.z.values = dimOneTile.z.values - dimOneTile.z.values(focusPositionInImageZpix(json.zDepths == 0));
+    if isempty(json.zDepths) || ~isnumeric(json.zDepths) % Validate presence of z-depths data
+        error('ScanInfo.json reports zero scans (empty zDepths). This volume appears to be empty or failed.');
+    end
+    [~, idx] = min(abs(json.zDepths)); % Adjust z-dimensions using the focus index at zDepth 0
```
MyYo, line 51: I think it will be better to pick zDepth==0 and not min(abs(json.zDepths))
min can be very large number :)

MyYo (changes requested):
Thanks, Small comment

Processing/yOCTProcessTiledScan_createDimStructure.m
MyYo: Or, make sure that json.zDepths is less than 10 microns and is the minimum.

alejoruge: Ohh yees, I'm sorry. That makes more sense. Resolved.

alejoruge:
Please let me know if you agree with checking 10 microns from 0 like this.

alejoruge:
Sorry about the double commit... The first one had a bad entry, so I had to re-commit. Would you prefer I open a new branch to keep it clean?

MyYo (approved):
Thanks, looks good!

PR #25, opened 2025-04-04
Update serialport

aidanvanvleck:
Updated serialport from COM4 to COM3 to work with new optical switch


MyYo (approved):
Looks good to me!

PR #30, opened 2025-04-17
Changing input units to yOCTFindTissueSurface will change the output

MyYo:
According to yOCTFindTissueSurface's documentation, output parameters: urfacePosition,x,y are in mm
However, when changing the dimensions of the input files the output values change.

In this pull request, I created a test to verify that output values are kept identical even if inputs change.
The test currently fails.

Diego, please:
1. Fix yOCTFindTissueSurface such that the test passes.
2. Commit the fix to this branch.
3. Assign me as the reviewer of the fix.

Thank you!


alejoruge:
Hi, thank you for the tester!

I wanted to mention the commented out part of the previous function in the tester, which was checking if the output would match the input units. However, with the newer version, the goal is actually the opposite I think, to ensure the output is always returned in millimeters.

Please take a look at the updated function, as I made sure it now consistently outputs in millimeters. I just wanted to double-check with you before permanently deleting the old tester function.

Thank you!

alejoruge (commented):
Hi, thank you for the tester!

I wanted to mention the commented out part of the previous function in the tester, which was checking if the output would match the input units. However, with the newer version, the goal is actually the opposite I think, to ensure the output is always returned in millimeters.

Please take a look at the updated function, as I made sure it now consistently outputs in millimeters. I just wanted to double-check with you before permanently deleting the old tester function.

Thank you!

alejoruge:
I opened PR #36 to continue the changes requested here, because this PR needed to be organized into atomic commits focused on the requirement and it wasn’t possible to assign reviewers.

This PR and its branch can now be deleted.

PR #31, opened 2025-04-17
Split finding tissue surface from asserting that tissue is in focus

MyYo:
Currently yOCTScanAndFindTissueSurface does two things:
1. Finds the focus
2. Asserts that the scan is in focus

The purpose of this script is to separate the two functionalities in order to create more atomic functions
yOCTScanAndFindTissueSurface will only find the focus.
yOCTAssertTissueSurfaceIsInFocus will assert that the tissue is in focus.

In this pull request I split the function, created a tester and modified Demo to include the assertion statement.


alejoruge (approved):
Thank you!

Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -59,22 +62,23 @@
...
     );
 
 if in.skipHardware
-    % No need to continue
-    surfacePosition_mm = 0;
-    x_mm = 0;
-    y_mm = 0;
-    return;
-end
+     % No need to continue
```
MyYo, line 73: I think there is extra space here

Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -29,22 +30,24 @@
...
 in = p.Results;
+
 xRange_mm = in.xRange_mm;
 yRange_mm = in.yRange_mm;
 pixel_size_um = in.pixel_size_um;
 isVisualize = in.isVisualize;
+octProbeFOV_mm = in.octProbeFOV_mm;
 octProbePath = in.octProbePath;
-output_folder = in.output_folder;
+output_folder = [in.output_folder '\surfaceAnalysis_temp'];
```
MyYo, line 50: Worth changing the name to "temporary_files_folder"

Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -105,23 +109,29 @@
...
+elapsedTimeSurfaceDetection_sec = toc(tSurface);
 if (v)
     fprintf('%s Surface identification completed in %.2f seconds.\n', datestr(datetime), elapsedTimeSurfaceDetection_sec);
 end
 
 %% Clean up
-if exist(output_folder, 'dir')
+if ~isempty(output_folder) && exist(output_folder, 'dir')
     rmdir(output_folder, 's'); % Remove the output directory after processing
+    output_folder = fileparts(output_folder); % Return to original output_folder
```
MyYo, line 130: Can you explain that? Why is this used?

MyYo (commented):
Thanks, a few corrections

Applications/yOCTScanAndFindTissueSurface.m
alejoruge: Oh well I just wanted to align it back with the original user input but there’s no real need for a change here, so I reverted it. Resolved.

PR #34, opened 2025-04-29
This is a refactor of yOCTPhotobleachTile to enable multiple z depths

MyYo:
It's a big PR so I will break it down.
yOCTPhotobleachTile had to be split to "make a plan" and "photobleach the plan" in two functions. This refactor allows us to photobleach at multiple depths.

In addition, I made a small change to yOCTProcessTiledScan.m - this one you can just approve, it was a bug on my end that I fixed.


ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -100,151 +100,24 @@
...
-    end
-    hold off;
-    axis equal;
-    axis ij; % Match coordinate system: https://docs.google.com/presentation/d/1tOod76WvhvOuByo-K81YB4b3QjRq-6A5j2ztS_ANSNo/edit#slide=id.g2812f1bd1d1_0_115
-    grid on;
-    xlabel('x[mm]');
-    ylabel('y[mm]');
-	title(sprintf('Photobleach Pattern\nEstimated Time: %.0f minutes',ceil(estimatedPhotobleachTime_sec/60)));
+    % Estimate photobleach time
+    totalLineLength = sum(sum([PI.lineLengths_mm{:}])); % mm
```
alejoruge, line 114: The variable PI doesn’t exist in the code. It’s a typo. I think you meant to use photobleachPlan, which contains the line lengths in the field lineLength_mm.

like this maybe? totalLineLength = sum([photobleachPlan.lineLength_mm]); % mm

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -309,31 +180,36 @@
 
 %% Photobleach pattern
 
-% Loop over FOVs
-for i=1:length(xcc)
+% Go over plan
+for i=1:length(photobleachPlan)
+    ppStep = photobleachPlan(i);
+
+    if (v && (length(xcc) > 1 || length(json.z) > 1) )
```
alejoruge, line 187: xcc isn’t defined, it’s probly also a typo or leftover from a previous version. The code is trying to check if there are multiple stage positions (xcc) or depths (json.z), but xcc doesn’t exist here

if (v && (length(photobleachPlan) > 1 || length(json.z) > 1) )

ThorlabsImager/yOCTPhotobleachTile.m
MyYo: You are correct.

ThorlabsImager/yOCTPhotobleachTile.m
MyYo: Updated.

PR #35, opened 2025-05-03
Small refactor, see description

MyYo:
1. If focusPositionInImageZpix is not provided, return an error. It's good engineering practice to not have user inputs inside a closed function. For this reason, if user didn't provide value, the function will fail.
2. Removing subfolders from output_folder
3. Removing user message " Please adjust the OCT focus such that it is precisely at the intersection of the tissue and the coverslip." because it's not true. The purpose of the function is to find the focus position, so we can't ask a user to provide it :)


alejoruge (commented):
Hi, thank you.

I agree with focus position part, thanks for the lesson, I think it will never be empty, it is either number or NaN, and in regards to the message (point 3) I totally agree, not really even sure why that was there..

In the second, '**Removing subfolders from output_folder**': so the reason I added the temporary subfolder was because the cleanup step deletes everything in the output folder, even files that were already there, and by accident I once lost a whole scan when the process saved it early and then cleanup wiped it. Considering that now the process is First surface scan and second OCT scan, the order's changed so that's less likely to happen. But I kept it because the cleanup still deletes any file in that folder and I was worried a user might have other important files in their chosen output folder that would get accidentally erased. So considering this what do you think?

PR #37, opened 2025-05-03
Add a tester for challenging cases from roboflow

MyYo:
We can't simulate everything, it's therefore good to have some data in roboflow with real life challenging cases and to verify algorithm performances on those cases.

Before merging this, lets talk about what few fist case to add to roboflow.


alejoruge:
This is great! I will post an update with next PR to ensure it works as intended. But amazing!

alejoruge:
Please see PR #44 for continuation of this tester.

PR #38, opened 2025-05-08
Fix bug with slice‑spacing (y axis) metadata in ImageJ/Fiji

alejoruge:
Problem
-------

- Fiji had slices always show with 1.00 as spacing in the Y axis (Z in Fiji), regardless of the pixel_size. This made 3D views look wrong and not true to the actual scale when pixel size is different to 1um. For instance, an x, y = 6mm x 6mm image appeared as 0.6mm x 6mm if pixel_size was 10 um.

- Warning was spamming the console when dimensions were not provided.

What changed:

* Validates Y‑axis metadata and computes pixelSizeY_um.
* Adds `spacing=<pixelSizeY_um>` to the ImageJ header string (Z axis in Fiji/each slide).
* Removes the loop‑level warning that was spamming the console.

Result:

- Fiji now shows the correct voxel y axis (z in Fiji), appearing at true scale in 3‑D viewer and Properties.
- No Warning spamming the console if no dimensions provided.
- All tests passed.

### DIFF
--- a/LoadSave/yOCT2Tif.m
+++ b/LoadSave/yOCT2Tif.m
@@ -211,12 +211,14 @@ if mode == 0
             % Check for valid metadata to set resolution accordingly in ImageJ
             if ~isempty(metadata) && ...
                 isfield(metadata, 'x') && isfield(metadata.x, 'values') && ...
+                isfield(metadata, 'y') && isfield(metadata.y, 'values') && ...
                 isfield(metadata, 'z') && isfield(metadata.z, 'values') && ...
-                numel(metadata.x.values) > 1 && numel(metadata.z.values) > 1
+                numel(metadata.x.values) > 1 && numel(metadata.y.values) > 1 && numel(metadata.z.values) > 1
 
                 % Convert dimension structure to microns for accurate pixel spacing in ImageJ
                 meta_um = yOCTChangeDimensionsStructureUnits(metadata,'microns');
                 pixelSizeX_um = abs(meta_um.x.values(2) - meta_um.x.values(1));
+                pixelSizeY_um = abs(meta_um.y.values(2) - meta_um.y.values(1));
                 pixelSizeZ_um = abs(meta_um.z.values(2) - meta_um.z.values(1));
 
                 % ImageJ uses 'XResolution', 'YResolution', 'ResolutionUnit', and reads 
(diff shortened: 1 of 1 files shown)


LoadSave/yOCT2Tif.m
```diff
@@ -211,23 +211,31 @@
             % Check for valid metadata to set resolution accordingly in ImageJ
             if ~isempty(metadata) && ...
                 isfield(metadata, 'x') && isfield(metadata.x, 'values') && ...
+                isfield(metadata, 'y') && isfield(metadata.y, 'values') && ...
                 isfield(metadata, 'z') && isfield(metadata.z, 'values') && ...
-                numel(metadata.x.values) > 1 && numel(metadata.z.values) > 1
+                numel(metadata.x.values) > 1 && numel(metadata.y.values) > 1 && numel(metadata.z.values) > 1
```
MyYo, line 216: Can you explain (in a comment) what it is you are checking? Are you checking that the length is more than 1?

LoadSave/yOCT2Tif.m
MyYo, line 1: Thanks. What will happen if this is a B scan (with size 0 in Y)? Will the dimensions be wrong now?

LoadSave/yOCT2Tif.m
```diff
@@ -211,23 +211,31 @@
...
                 % ImageJ uses 'XResolution', 'YResolution', 'ResolutionUnit', and reads 
                 % 'ImageDescription' (e.g., "unit=um") to handle units and scaling display
                 tagstruct.XResolution         = 1/(pixelSizeX_um*1e-4); % Resolution in microns
                 tagstruct.YResolution         = 1/(pixelSizeZ_um*1e-4); % Z is Y in ImageJ/Fiji
                 tagstruct.ResolutionUnit      = Tiff.ResolutionUnit.Centimeter; % Inch also possible
-                tagstruct.ImageDescription    = sprintf('ImageJ=1.53\nunit=um\nspacing=1.00\nimages=%d\n', size(data,3));
+                tagstruct.ImageDescription = sprintf( ...
+                    ['ImageJ=1.53\n' ...
+                     'unit=um\n' ...
+                     'spacing=%g\n' ...         % Spacing is Y axis (Z in Fiji)
```
MyYo, line 232: Can you explain this: "Spacing is Y axis (Z in Fiji)"

Do you mean: "Set Z spacing in Fiji which is the same as Y axis in metadata?

LoadSave/yOCT2Tif.m
```diff
@@ -211,23 +211,31 @@
...
+                     'images=%d\n'], ...
+                     pixelSizeY_um, size(data,3));
             else
+
                 % Default to 1 px if metadata is invalid or absent
-                warning('No valid metadata provided. Using default resolution = 1 pixel (no real-world scaling).');
+                % No valid metadata provided. Using default resolution = 1 pixel (no real-world scaling).
```
MyYo, line 238: Missing z resolution here no?

MyYo (changes requested):
Thanks, a few changes needed

LoadSave/yOCT2Tif.m
alejoruge: Yes I agree, so default slice spacing is already 1.00 px when the header omits it, but I’ve added spacing=1.00 explicitly, i think it's a good practice so the assumption is visible in the metadata as well and for anyone reading it. Resolved.

LoadSave/yOCT2Tif.m
alejoruge: Right! Resolved.

LoadSave/yOCT2Tif.m
alejoruge: Yes, if there's only one B-scan, then we can’t determine the pixel size along the Y axis, because we need at least two Y positions to calculate the spacing (we use the difference between the first and second values).

In theory, we could assume it's the same as the X spacing just to fill in a value, but I'm not sure if that's a good practice... not based on real data, and might lead to misleading visualizations and info

So instead, the current logic skips adding spacing when we don’t have enough information, which I thought was the safest and cleanest option, but maybe there's another way to get the pixel_size? 

We could also add this information it in the metadata.. I'm not sure please let me know your take

LoadSave/yOCT2Tif.m
alejoruge: Yes, that's correct. I added this comment, please let me know what you think:

            % Proceed only if metadata contains at least TWO samples in each axis (z, x, y).
            % We need ≥2 points to compute pixel spacing => pixelSize = |v(2) – v(1)|.
            % If any axis has ≤1 value we skip physical scaling and fall back to 1‑pixel units

LoadSave/yOCT2Tif.m
MyYo: got it okay.

MyYo (approved):
Looks good.
Thanks!

PR #41, opened 2025-05-15
Auto‑adjust focus to tissue‑coverslip interface within OCT FOV

alejoruge:
The scan now checks focus only inside the square actually captured by the OCT scan FOV.

If the surface is too high or too low, the script moves the Z‑stage by that exact amount and shows a warning. The scan continues and all data stay in memory.

Added cropToFOV helper to extract (or, if smaller than the data, keep) the exact area of the main OCT scan before judging focus, not a larger one.

Users see a yellow warning when the Z move happens & extra numbers appear when verbose is on.

Result: the main OCT scan is guaranteed to be in focus, without stopping the run.

### DIFF
--- a/Applications/yOCTScanAndFindTissueSurface.m
+++ b/Applications/yOCTScanAndFindTissueSurface.m
@@ -138,9 +138,46 @@ if (v)
         datestr(datetime), datestr(totalDuration, 'HH:MM:SS'));
 end
 
-%% Assert
+%% Assert focus is within acceptable range
 if ~isempty(in.assertInFocusAcceptableRange_mm)
-    yOCTAssertTissueSurfaceIsInFocus( ...
-        surfacePosition_mm, x_mm, y_mm, ...
-        in.assertInFocusAcceptableRange_mm, v);
+    % Get FOV (use probe.ini if [] was passed)
+    if isempty(octProbeFOV_mm)
+        tmpIni = yOCTReadProbeIniToStruct(octProbePath);
+        octProbeFOV_mm = tmpIni.RangeMaxX;    % lens X range
+    end
+    % Crop surface map to that FOV
+    [surfFOV_mm, xFOV_mm, yFOV_mm] = cropToFOV( ...
+        surfacePosition_mm, x_mm, y_mm, octProbeFOV_mm );
+    % Run the assertion inside TRY‑CATCH so we don't lose surface data
+    if numel(surfFOV_mm) > 0      % only assert if crop not empty
+         try
+            yOCTAssertTissueSurfaceIsInFocus( ...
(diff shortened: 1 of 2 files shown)


Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -138,9 +138,46 @@
         datestr(datetime), datestr(totalDuration, 'HH:MM:SS'));
 end
 
-%% Assert
+%% Assert focus is within acceptable range
 if ~isempty(in.assertInFocusAcceptableRange_mm)
-    yOCTAssertTissueSurfaceIsInFocus( ...
-        surfacePosition_mm, x_mm, y_mm, ...
-        in.assertInFocusAcceptableRange_mm, v);
+    % Get FOV (use probe.ini if [] was passed)
```
MyYo, line 143: Instead of this implementation, I recommend providing an input parameter "assertInFocusAcceptableRangeXYArea_mm" that user explicitly provides what is the region of interest to be in focus. If user provides nothing, assertion would work on the entire area.

Processing/yOCTAssertTissueSurfaceIsInFocus.m
```diff
@@ -27,20 +29,33 @@ function yOCTAssertTissueSurfaceIsInFocus( ...
...
 if abs(averageSurfaceDistance_mm) > acceptableRange_mm
     if averageSurfaceDistance_mm > 0 % Determine direction of adjustment
-        direction = 'increase';
+        direction = 'increase';   % surface is below focus = raise stage
     else
-        direction = 'decrease';
+        direction = 'decrease';   % surface is above focus = lower stage
     end
+
+    % Automatic Z correction
```
MyYo, line 40: Move the physical movement part to 
"yOCTScanAndFindTissueSurface.m"
Make sure to move only if
skipHardware = false
Add a parameter "moveTissueToFocusIfNeeded" (default: true) that determines if movement will be done

MyYo (changes requested):
Thanks, left a few comments.

FYI, it seems that the code that was pasted from chatGPT has some hidden charecters. Make sure to remove them (GitHub's error):
The head ref may contain hidden characters: "assert\u2011focus\u2011auto\u2011adjust"

Applications/yOCTScanAndFindTissueSurface.m
alejoruge: Resolved

Processing/yOCTAssertTissueSurfaceIsInFocus.m
alejoruge: Resolved

PR #43, opened 2025-05-21
Quick Fix: ensure process function uses correct pixel size

alejoruge:
Passed **outputFilePixelSize_um** to avoid assert failure in **yOCTProcessTiledScan** **(line 194)** when pixelSize_um ≠ 1

Renamed pixel_size_um → pixelSize_um for style consistency

See pic of error when pixelSize_um of scan was 5, and yOCTProcessTiledScan takes the default 1:

pixelSize_um_error


MyYo (approved):
Good catch! Thank you

PR #44, opened 2025-05-23
Support real-world OCT image testing with clim-preserving PNG transfo…

alejoruge:
This update ensures surface detection works with real OCT data, even when running through Roboflow's JPG coming from PNG.

Roboflow doesn’t support raw .tif files, so we built a conversion layer that:
• Turns OCT TIFF/TIFs into PNGs that Roboflow can read
• Stores the original scale (clim) in the filename
• Recovers that data after JPG export from Roboflow

Now surface detection code can work without changes, and testing on OCT images. Also includes a visual selector for 3D OCT volumes for when preparing samples to be used for the tester.

It does:

• Updated the Roboflow tester to use 20 real-world OCT images (dataset v4).

• Fixed small bug in finding surface pixel.

• Introduced a new helper:

yOCTConvertAndReadTif2Png

WHY:

Surface detection requires original OCT intensity values, but Roboflow forces PNG/JPG, which destroys them. This essentially breaks the detection code.

SOLUTION:

We now convert real OCT .tif / .tiff files into .png while saving their intensity range (clim) in the filename. Later, when Roboflow sends back JPGs, we restore the original values using that info.

- Handles multi-slice TIFFs with a visual slice picker to prepare new scan images in PNG for testing.
- Fully compatible with surface detection code.

### DIFF
--- a/Processing/test_yOCTFindTissueSurface.m
+++ b/Processing/test_yOCTFindTissueSurface.m
@@ -145,14 +145,14 @@ classdef test_yOCTFindTissueSurface < matlab.unittest.TestCase
             % and tests that they pass
 
             % Load dataset
-            rfFolder = rfDownloadDataset('tissue-surface','outputFolder','./temp/');
+            rfFolder = rfDownloadDataset('tissue-surface','outputFolder','./temp/', 'version', 4);
             [imageFilePath, imageMaskPath, ~] = ...
                 rfListImagesAndSegmentationsInDataset(rfFolder,'above-tissue');
 
             % Loop over all images and compare algorithm with ground truth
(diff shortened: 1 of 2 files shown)


alejoruge:
Using this helper function (with no modifications to the surface detection function), all 20 datasets in version 4 passed. The helper can be further adjusted based on feedback. Please see the results below with this PR and Roboflow v4:

v4_testResults

MyYo (changes requested):

alejoruge:
I closed this PR. Just provided alternative to local folder for the test with original TIFF/TIF files and their PNG masks here: [PR](https://github.com/WinetraubLab/myOCT/pull/47)

PR #45, opened 2025-05-29
Use yOCTAssertTissueSurfaceIsInFocus with error handling to move Z-stage if needed

alejoruge:
Integrated yOCTAssertTissueSurfaceIsInFocus with structured error handling as per PR #42 

Move Z-stage if assert triggers SurfaceOutOfFocus, all other checks pass, and hardware is enabled.

**Note: This continues the workk from PR #41 which had branch name issues and compatibility now here with main resolved issues.

### DIFF
--- a/Applications/yOCTScanAndFindTissueSurface.m
+++ b/Applications/yOCTScanAndFindTissueSurface.m
@@ -15,6 +15,18 @@ function [surfacePosition_mm, x_mm, y_mm] = yOCTScanAndFindTissueSurface(varargi
 %       set to [] to skip assertion.
 %   v: Verbose mode for debugging purposes and visualization default is 
 %       false.
+%   assertInFocusAcceptableRangeXYArea_mm : XY area (in mm) where the tissue 
+%       surface must be in focus. This defines the region used to check 
+%       whether the surface is within the assertInFocusAcceptableRange_mm 
+%       to the focus plane. If it's out of range, the Z stage will be
+%       automatically adjusted to bring the tissue into focus via
+%       yOCTAssertTissueSurfaceIsInFocus. Accepted values:
+%         • []             → uses the entire scan area (default)
+%         • single number  → centered square. Example 0.5 → –0.25 to +0.25 mm
+%         • 4 numbers      → [xMin xMax yMin yMax] custom rectangle (mm)
+%   moveTissueToFocusIfNeeded: (Default = true) If skipHardware is false and
+%       the surface is out of focus, having this true automatically moves the 
+%       Z stage by the required amount. False will skip this movement.
 %   skipHardware: Set to true to skip hardware operation. Default: false.
 % OUTPUTS:
 %   - surfacePosition_mm - 2D matrix. dimensions are (y,x). What
@@ -34,26 +46,40 @@ addParameter(p,'octProbePath','probe.ini',@ischar);
 addParameter(p,'temporaryFolder','./SurfaceAnalysisTemp/');
 addParameter(p,'dispersionQuadraticTerm',79430000,@isnumeric);
 addParameter(p,'focusPositionInImageZpix',NaN,@isnumeric);
-addParameter(p,'assertInFocusAcceptableRange_mm',0.025)
+addParameter(p,'assertInFocusAcceptableRange_mm',0.025);
+addParameter(p,'assertInFocusAcceptableRangeXYArea_mm',[], ... 
+    @(x) isempty(x) || isnumeric(x));
+addParameter(p,'moveTissueToFocusIfNeeded',true,@islogical);
 addParameter(p,'v',false);
 addParameter(p,'skipHardware',false)
 
 parse(p,varargin{:});
 in = p.Results;
 
-xRange_mm = in.xRange_mm;
-yRange_mm = in.yRange_mm;
-pixelSize_um = in.pixelSize_um;
-octProbeFOV_mm = in.octProbeFOV_mm;
-octProbePath = in.octProbePath;
+xRange_mm               = in.xRange_mm;
+yRange_mm               = in.yRange_mm;
+pixelSize_um            = in.pixelSize_um;
+octProbeFOV_mm          = in.octProbeFOV_mm;
+octProbePath            = in.octProbePath;
 dispersionQuadraticTerm = in.dispersionQuadraticTerm;
-temporaryFolder = in.temporaryFolder;
-v = in.v;
+temporaryFolder         = in.temporaryFolder;
+v                       = in.v;
 
 if isnan(in.focusPositionInImageZpix)
     error('Please provide a valid "focusPositionInImageZpix". Use yOCTFindFocusTilledScan to estimate.');
 end
 focusPositionInImageZpix = in.focusPositionInImageZpix;
+
+roi = in.assertInFocusAcceptableRangeXYArea_mm; % early roi validation
+if ~isempty(roi)
+    if isscalar(roi)
+        if roi <= 0
+            error('assertInFocusAcceptableRangeXYArea_mm scalar value must be positive (width in mm).');
+        end
+    elseif ~(numel(roi)==4)
+        error('assertInFocusAcceptableRangeXYArea_mm must be [] , a positive single number, or a 4‑element vector [xMin xMax yMin yMax].');
+    end
+end
 
 %% Scan
 totalStartTime = datetime;  % Capture the starting time
@@ -65,11 +91,11 @@ yOCTScanTile (...
     volumeOutputFolder, ...
     xRange_mm, ...
     yRange_mm, ...
-    'octProbeFOV_mm', octProbeFOV_mm, ...
-    'octProbePath', octProbePath, ...
-    'pixelSize_um', pixelSize_um, ...
-    'v',v,  ...
-    'skipHardware', in.skipHardware ...
+    'octProbeFOV_mm',  octProbeFOV_mm, ...
+    'octProbePath',    octProbePath, ...
+    'pixelSize_um',    pixelSize_um, ...
+    'v',               v,  ...
+    'skipHardware',    in.skipHardware ...
     );
 
 if in.skipHardware % No need to continue
@@ -87,10 +113,11 @@ outputTiffFile = [temporaryFolder '\surface_analysis.tiff'];
 yOCTProcessTiledScan(...
     volumeOutputFolder, ... Input
     {outputTiffFile},... Save only Tiff file as folder will be generated after smoothing
-    'focusPositionInImageZpix',focusPositionInImageZpix,...
-    'dispersionQuadraticTerm',dispersionQuadraticTerm,...
-    'cropZAroundFocusArea', false,...
-    'v',v);
+    'focusPositionInImageZpix', focusPositionInImageZpix, ...
+    'dispersionQuadraticTerm',  dispersionQuadraticTerm, ...
+    'outputFilePixelSize_um',   pixelSize_um,...
+    'cropZAroundFocusArea',     false, ...
+    'v',                        v);
 [logMeanAbs, dimensions, ~] = yOCTFromTif(outputTiffFile);
 dimensions = yOCTChangeDimensionsStructureUnits(dimensions,'millimeters'); % Make sure dimensions in mm
 if (v)
@@ -115,7 +142,7 @@ if (v)
     fprintf('%s Surface identification completed in %.2f seconds.\n', datestr(datetime), elapsedTimeSurfaceDetection_sec);
 end
 
-%% Clean up
+%% Clean up temp data
 if ~isempty(temporaryFolder) && exist(temporaryFolder, 'dir')
     rmdir(temporaryFolder, 's'); % Remove the output directory after processing
 end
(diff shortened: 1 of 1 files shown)


Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -15,6 +15,18 @@
 %       set to [] to skip assertion.
 %   v: Verbose mode for debugging purposes and visualization default is 
 %       false.
+%   assertInFocusAcceptableRangeXYArea_mm : XY area (in mm) where the tissue 
+%       surface must be in focus. This defines the region used to check 
+%       whether the surface is within the assertInFocusAcceptableRange_mm 
+%       to the focus plane. If it's out of range, the Z stage will be
+%       automatically adjusted to bring the tissue into focus via
+%       yOCTAssertTissueSurfaceIsInFocus. Accepted values:
```
MyYo, line 23: Although I agree giving options to user is nice. In this case, I think it will create more confusion then help.
Which of these options are actually being used today? I think you should keep only the ones that we are currently using maybe empty and four numbers?

Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -128,7 +155,85 @@
 
 %% Assert
 if ~isempty(in.assertInFocusAcceptableRange_mm)
-    yOCTAssertTissueSurfaceIsInFocus( ...
-        surfacePosition_mm, x_mm, y_mm, ...
-        in.assertInFocusAcceptableRange_mm, v);
+    % Decide which XY pixels to include (Region of Interest)
+    if isempty(roi)          % [] -> keep full area
```
MyYo, line 159: Rows 159 - 180 are more suitable inside "yOCTAssertTissueSurfaceIsInFocus" because they are features of the assertion.

Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -128,7 +155,85 @@
...
+                criticalFail = true;           % never move Z stage in these cases
+            case 'yOCT:SurfaceOutOfFocus'
+                outOfFocusErr = true;          % we MAY move Z stage
+            otherwise
+                % unexpected error -> treat as critical
+                criticalFail = true;
+        end
+    end
+
+    roiVals = surfROI_mm(:);
```
MyYo, line 201: Rows 201 - 211 need to be inside yOCTAssertTissueSurfaceIsInFocus, yOCTAssertTissueSurfaceIsInFocus is providing an estimate to the user of how much moving is needed. So aren't you recalculating it here?

Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -128,7 +155,85 @@
...
+                ~isnan(medianSurfacePosition_mm)                                    && ...  
+                abs(medianSurfacePosition_mm) > in.assertInFocusAcceptableRange_mm  && ...  
+                ~in.skipHardware                                                    && ...
+                in.moveTissueToFocusIfNeeded;
+
+    if needMove
+        [~,~,z0] = yOCTStageInit();  % query current Z
+        try
+            yOCTStageMoveTo(NaN, NaN, z0 + medianSurfacePosition_mm, v);
+            warning('%s Stage auto‑moved by %.3f mm to refocus tissue surface.', ...
```
MyYo, line 225: Why warning? is this something that the user should be concerned? fix?

Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -128,7 +155,85 @@
...
+
+            % keep surface map consistent with new focus
+            surfacePosition_mm = surfacePosition_mm - medianSurfacePosition_mm;
+
+            if v
+                fprintf('%s Stage Z successfully MOVED from %.3f mm to %.3f mm (OCT coord).\n', ...
+                        datestr(datetime), z0, z0 + medianSurfacePosition_mm);
+            end
+        catch ME
+            warning(ME.identifier, 'Stage move failed: %s', ME.message);
```
MyYo, line 236: Why isn't that an error?

Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -128,7 +155,85 @@
...
+
+        surfROI_mm = surfacePosition_mm(iy,ix);
+        xROI_mm    = x_mm(ix);
+        yROI_mm    = y_mm(iy);
+    end
+
+    criticalFail  = false;
+    outOfFocusErr = false;
+
+    try
```
MyYo, line 185: This implementation seems inefficient. Can you explain what you are trying to do here? Why not leave the assert outside a try catch?

MyYo (changes requested):
Thanks Diego, here are some comments.

Applications/yOCTScanAndFindTissueSurface.m
alejoruge: Right, moved them to assert. Ensure the provided information is passed.

Applications/yOCTScanAndFindTissueSurface.m
alejoruge: I thought yes, since the Z stage will be moving and user should be be aware somehow that they put it wrong so it doesn't repeat, no? Because that might also have been a wrong input (like wrong focus), and I thought it would be valuable if user knows for sure what happened. I printed in this new version instead I think showing could be beneficial. Please tell me what you think.

Applications/yOCTScanAndFindTissueSurface.m
alejoruge: So the assert inside a try catch because it throws an error if **yOCT:surfaceCannotBeEstimated**, or **yOCT:SurfaceCannotBeInFocus**. But, if error is **yOCT:SurfaceOutOfFocus**, then we will attempt to move if user wanted to, so I thought catching the error will help identifying which scenario applies. Please let me know your thoughts.

Processing/yOCTAssertTissueSurfaceIsInFocus.m
```diff
@@ -1,5 +1,5 @@
-function yOCTAssertTissueSurfaceIsInFocus( ...
-    surfacePosition_mm, x_mm, y_mm, acceptableRange_mm)
+function medianSurfacePosition_mm = yOCTAssertTissueSurfaceIsInFocus( ...
```
MyYo, line 1: Change to
[isSurfaceInFocus, surfacePosition_mm]

isSurfaceInFocus = true - in focus
isSurfaceInFocus = false out of focus

surfacePositionOutput_mm = representative focus position of the checked area

Processing/yOCTAssertTissueSurfaceIsInFocus.m
```diff
@@ -10,6 +10,12 @@ function yOCTAssertTissueSurfaceIsInFocus( ...
 %   y_mm: provided by yOCTScanAndFindTissueSurface
 %   acceptableRange_mm: how far can tissue surface be from focus position
 %       to be considered "good enough". Default: 0.025mm
+%   assertInFocusAcceptableRangeXYArea_mm : XY area (in mm) where the tissue
```
MyYo, line 13: Have an input
throwErrorIfOutOfFocus, default is true

MyYo: assertInFocusAcceptableRangeXYArea_mm change to 
roiToCheckSurfacePosition = ROI defined as [x,y,width,height] in mm
Keep empty to use all area.

Processing/yOCTAssertTissueSurfaceIsInFocus.m
```diff
@@ -10,6 +10,12 @@ function yOCTAssertTissueSurfaceIsInFocus( ...
 %   y_mm: provided by yOCTScanAndFindTissueSurface
 %   acceptableRange_mm: how far can tissue surface be from focus position
 %       to be considered "good enough". Default: 0.025mm
+%   assertInFocusAcceptableRangeXYArea_mm : XY area (in mm) where the tissue 
+%       surface must be in focus. This defines the region used to check 
+%       whether the surface is within the assertInFocusAcceptableRange_mm 
+%       to the focus plane. Accepted values:
+%         [] (default)   =  uses the entire scan area
```
MyYo, line 17: Default [] checks all area, otherwise should be a four element vector [x,y,width, height] mm

Processing/yOCTAssertTissueSurfaceIsInFocus.m
```diff
@@ -53,4 +75,4 @@ function yOCTAssertTissueSurfaceIsInFocus( ...
     error(errorID, msg); 
 end
 
-fprintf('%s The average distance of the surface (%.3f mm) is within the acceptable range.\n', datestr(datetime), medianSurfacePosition_mm);
\ No newline at end of file
+fprintf('%s The average distance of the surface (%.3f mm) is within the acceptable range.\n', datestr(datetime), medianSurfacePosition_mm);
```
MyYo, line 78: if v?

Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -21,9 +21,8 @@
 %       to the focus plane. If it's out of range, the Z stage will be
 %       automatically adjusted to bring the tissue into focus via
 %       yOCTAssertTissueSurfaceIsInFocus. Accepted values:
-%         • []             → uses the entire scan area (default)
-%         • single number  → centered square. Example 0.5 → –0.25 to +0.25 mm
-%         • 4 numbers      → [xMin xMax yMin yMax] custom rectangle (mm)
+%         [] (default)   =  uses the entire scan area (default)
+%         single number  =  centered square area. Example 0.5 makes it –0.25 to +0.25 mm
 %   moveTissueToFocusIfNeeded: (Default = true) If skipHardware is false and
```
MyYo, line 26: Change name to "moveTissueToFocus". Move tissue to focus automatically if tissue is out of focus. If Skiphardware = true, moving is disabled.

Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -72,12 +71,8 @@
 
 roi = in.assertInFocusAcceptableRangeXYArea_mm; % early roi validation
 if ~isempty(roi)
-    if isscalar(roi)
-        if roi <= 0
-            error('assertInFocusAcceptableRangeXYArea_mm scalar value must be positive (width in mm).');
-        end
-    elseif ~(numel(roi)==4)
-        error('assertInFocusAcceptableRangeXYArea_mm must be [] , a positive single number, or a 4‑element vector [xMin xMax yMin yMax].');
+    if ~(isscalar(roi) && roi>0)
```
MyYo, line 74: This is already checked in line input

Applications/yOCTScanAndFindTissueSurface.m
```diff
@@ -142,87 +137,48 @@
...
-    if isempty(roiVals)
-        medianSurfacePosition_mm = NaN;
-    else
-        maxStageShift_mm = 0.20; % safety cap for maximum Z move (200 um)
-        medianSurfacePosition_mm = median(roiVals);
-        if abs(medianSurfacePosition_mm) > maxStageShift_mm
-            medianSurfacePosition_mm = sign(medianSurfacePosition_mm) * maxStageShift_mm;
-        end
+    
+    maxStageShift_mm = 0.20;            % 200 micron safety cap
```
MyYo, line 164: maxDownMovementSafetyCap_mm =0.2

MyYo (changes requested):

Applications/yOCTScanTissueSurfaceAndAutoAdjustFocusViaZStage.m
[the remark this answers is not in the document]
alejoruge: What do you think: **yOCTScanTissueSurfaceAndAutoAdjustFocus** or **yOCTScanTissueSurfaceAndAutofocus**

Demo_ScanSampleOverview.m
```diff
@@ -38,9 +39,10 @@
 %% Scan Overview
 
 % Quick pre-scan to identify tissue surface and verify it is at OCT focus
-[surfacePosition_mm, x_mm, y_mm] = yOCTScanAndFindTissueSurface(... 
-        'xRange_mm', sampleSize_mm*[-1 1],...
-        'yRange_mm', sampleSize_mm*[-1 1],...
+[surfacePosition_mm, x_mm, y_mm] = yOCTTissueSurfaceAutofocus(... 
+        'xRange_mm', sampleSize_mm*[-octProbeFOV_mm octProbeFOV_mm],...
```
MyYo, line 43: My intent was to scan sampleSize_mm. so 2mm by 2mm.
In this case you will be scanning 1mm by 1mm. It's not my intention

Applications/yOCTScanTissueSurfaceAndAutoAdjustFocusViaZStage.m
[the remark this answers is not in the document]
alejoruge: I was having some trouble making the changes directly here, so I created PR #53 as it seemed to be the cleanest option for review the changes. In all my other attempts, the diffs were showing up as full file additions and deletions..

Please take a look at that PR for this changes

PR #48, opened 2025-06-27
Correct mall bug with outputFilePixelSize_um

alejoruge:
If we pass [] (empty) for outputFilePixelSize_um, the code currently crashes.

Instead, it should skip the isotropy check and keep whatever native step sizes came from the scan (XY could be 10 µm, Z = ~1.44 microns).

Now if we need (like yOCTScanAndFindTissueSurface currently does), we will pass [] to get the original inputs.

Also quick bug with newer MATLAB versions where in a vector, first value must be explicit (line 200).

### DIFF
--- a/Processing/test_yOCTProcessTiledScan.m
+++ b/Processing/test_yOCTProcessTiledScan.m
@@ -1,0 +1,84 @@
+classdef test_yOCTProcessTiledScan < matlab.unittest.TestCase
+
+    properties
+
+    end
+    
+    methods(Test)
+        function testLoadSaveNoStitchingNoFocus(testCase)
+            % Confirm that yOCTProcessTiledScan works when:
+            %   outputFilePixelSize_um = [] -> keeps native spacing
+            %   outputFilePixelSize_um omitted (default 1 micron isotropic)
+            %   And that the Z-dimension length scales as expected.
+
+            % Generate Data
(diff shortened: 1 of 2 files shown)


alejoruge:
One quick question now that I’m touching on this point:

Right now, if the scan was done at 10 um laterally but we ask for 1 um in outputFilePixelSize_um (or don't pass anything since the default is the same 1 um), the code stops with this assert:

“pixelSizeX used for scanning is not the same as ‘outputFilePixelSize_um’, this is not implemented yet…”

Would it make sense to support XY scaling as well, so the final stack in this case is always 1 um × 1 um × 1 um regardless of the original scan step?

So just saying, if you think the isotropic 1 um output is worth it, I think I can prototype this by adjusting/stretching dimOutput_mm.x.values and .y.values when outputFilePixelSize_um is smaller than the scan’s spacing

Please let me know if this seems worth implementing at all!

MyYo (changes requested):
Thanks for the fix. It looks good
However, do you mind creating a tester that would demonstrate the failure of the bug you just fixed?

Create a tester class: test_yOCTProcessTiledScan in Processing folder

The tester class should use yOCTSimulateTileScan to create simulated data and then test yOCTProcessTiledScan in a way that will need this fix to pass.

MyYo (approved):
Thanks, looks good

PR #50, opened 2025-07-02
Switch surface confirmation check to microns & remove debug message

alejoruge:
Replaced fixed pixel-based surface confirmation threshold with microns-based threshol, more accurate since output pixel size can vary.

Also removed an unnecessary print statement.

Added tester


MyYo (approved):
Looks good thank you

PR #51, opened 2025-07-02
Set high_z_start to 100 pixels + image added to tester

alejoruge:
Set high_z_start to 100 for deep Z-stacks compatibility and add real world test image to the testRealWorldDataset


MyYo (approved):
Thanks for the fast turn around

PR #57, opened 2025-07-24
Add helper yOCTComputeZOffsetToFocusFromSurfaceROI for Z offset calculation

alejoruge:
- Added helper function to compute average tissue surface position within ROI and return Z offset for focus correction.
- Used by yOCTPhotobleachTile to adjust the Z-stage per tile during photobleaching (updated its use)
- Also used by yOCTAssertFocusAndComputeZOffset to validate focus and assist in surface assertions, including the ROI used.


Processing/yOCTAssertFocusAndComputeZOffset.m
```diff
@@ -17,12 +17,14 @@
...
 % OUTPUTS:
-%   isSurfaceInFocus: Resulting evaluation indicating whether the tissue surface is within 
-%       the acceptable range of the current focus position. Returns true if in focus, false otherwise.
 %   zOffsetCorrection_mm: Average Z distance between the current focus and the estimated 
 %       tissue surface in the specified ROI. It can be used to move the Z stage and 
 %       bring the tissue into focus. A positive value means the stage needs to move up;
 %       a negative value means it should move down.
+%   isSurfaceInFocus: Resulting evaluation indicating whether the tissue surface is within 
+%       the acceptable range of the current focus position. Returns true if in focus, false otherwise.
+%   roiSurfaceMap_mm: Cropped surface map (same dimensions as the selected ROI) returned as a 2D matrix.
```
MyYo, line 26: Try improving the documentation. This is not clear.

Processing/yOCTAssertFocusAndComputeZOffset.m
```diff
@@ -44,11 +46,23 @@
 assert(size(surfacePosition_mm,2) == length(x_mm),'surfacePosition_mm second dimension should match x_mm')
 
 % Compute Z offset correction to bring tissue into focus and extract ROI from the surface map
-[zOffsetCorrection_mm, roiSurfaceMap_mm] = yOCTComputeZOffsetToFocusFromSurfaceROI( ...
-        surfacePosition_mm, x_mm, y_mm, ...
-        'roiToCheckSurfacePosition', roiToCheck, ...
-        'v', v);
-roiSurfaceVec_mm = roiSurfaceMap_mm(:); % flatten surface map for numeric checks
+if isempty(roiToCheck)  % [] -> keep full area
```
MyYo, line 49: Can you add more comments. It's not clear what this part is doing

alejoruge:
The first of the two commits only merges recent changes from big_surface_update. The actual updates are in the most recent commit.

Regarding visualization here (and using the same example we discussed previously in this thread), with the same input data, two tiles would not have been photobleached here. The image below shows the resulting visualization, with the two affected tiles marked by red boxes and Xs. Please let me know your thoughts:

2_vetted_tiles

MyYo (changes requested):
Thanks. It may have been too long, I don't remember what is the goal of this commit. But here are some comments.

MyYo:
My recommendation is instead of doing this "
The first of the two commits only merges recent changes from big_surface_update. The actual updates are in the most recent commit."

Just add additional merge requests branching out of big_surface_update

MyYo (approved):

PR #58, opened 2025-07-31
Apply uniform Z offset to all tiles for consistent focal plane

alejoruge:
- Added support for uniformSurfaceOffset (default true) After computing per-tile Z offsets, all tiles now adopt the reference offset (the one identified for the scan 0,0)
- Stage Z is compensated so the laser focus remains at the same physical depth
- Prints the uniform offset in verbose mode

### DIFF
--- a/ThorlabsImager/yOCTPhotobleachTile.m
+++ b/ThorlabsImager/yOCTPhotobleachTile.m
@@ -27,6 +27,9 @@ function json = yOCTPhotobleachTile(varargin)
 %                                                   surfaceMap.surfacePosition_mm = surfacePosition_mm;
 %                                                   surfaceMap.surfaceX_mm        = surfaceX_mm;
 %                                                   surfaceMap.surfaceY_mm        = surfaceY_mm;
+%   uniformSurfaceOffset    true             If true, every tile is given the exact same zOffsetDueToTissueSurface taken from a single
+%                                            reference tile (central one at X=0 ,Y=0 or the one closest to it). This keeps the focal plane 
+%                                            uniform across the whole mosaic.
 %   exposure                15              How much time to expose each spot to laser light. Units sec/mm 
 %                                           Meaning for each 1mm, we will expose for exposurePerLine sec 
 %                                           If scanning at multiple depths, exposure will for each depth. Meaning two depths will be exposed twice as much. 
@@ -76,6 +79,7 @@ addParameter(p,'plotPattern',false);
 addParameter(p,'laserToggleMethod','OpticalSwitch');
 addParameter(p,'surfaceMap', [], @(x) isempty(x) || (isstruct(x) && ...
     all(ismember({'surfacePosition_mm','surfaceX_mm','surfaceY_mm'}, fieldnames(x)))));
+addParameter(p,'uniformSurfaceOffset',true,@islogical);
 
 parse(p,varargin{:});
 json = p.Results;
@@ -169,6 +173,41 @@ function photobleachPlan = adjustPlanZ(S, photobleachPlan)
     end
 end
 photobleachPlan = adjustPlanZ(json.surfaceMap, photobleachPlan);
+
+function photobleachPlan = applyUniformSurfaceOffset(photobleachPlan, json, v)
+% This function makes all tiles share one tissue-surface Z offset (identified
+% from the central scan). It is executed only when uniformSurfaceOffset == true
+
+    if ~json.uniformSurfaceOffset
+        return;                     % Keep original per-tile offsets
+    end
+
+    % Find the tile whose stageCenter (X,Y) is closest to (0,0).           
+    % That tile supplies the reference offset zRef.
+    [~,idx0] = min( hypot([photobleachPlan.stageCenterX_mm], ...
+                           [photobleachPlan.stageCenterY_mm]) ); % Locate reference tile (closest to 0,0)
+    zRef = photobleachPlan(idx0).zOffsetDueToTissueSurface; % Reference Offset
+    
+    % Propagate reference offset to all tiles to be photobleached
+    for k = 1:numel(photobleachPlan)
+        if k == idx0
+            continue; % it's already correct here
+        end
+        
+        % Apply the applied offsets
+        dZ = zRef - photobleachPlan(k).zOffsetDueToTissueSurface;
+        photobleachPlan(k).zOffsetDueToTissueSurface = zRef;
+        photobleachPlan(k).stageCenterZ_mm           = photobleachPlan(k).stageCenterZ_mm + dZ;
+    end
+
+    if v  % Print adjustments in verbose mode
+        fprintf('%s applyUniformSurfaceOffset: uniform zOffset used = %.4f mm\n', ...
+                datestr(datetime), zRef);
+    end
+end
+
+% Apply Uniform Z Offset if requested
+photobleachPlan = applyUniformSurfaceOffset(photobleachPlan, json, v);
 
 % Save the plan in the json
 json.photobleachPlan = photobleachPlan;
@@ -361,4 +400,3 @@ if (v)
 end
 end
 end
-


MyYo (changes requested):
Thanks. There are some refactor changes that are needed before I can continue review.

ThorlabsImager/test_yOCTPhotobleachTile_createPlan.m
```diff
@@ -344,38 +344,156 @@ function testUniformSurfaceOffset(tc)
                 ptStart = [-2 -2; -2  2];              % big X (start points)
                 ptEnd   = [ 2  2;  2 -2];              % big X (end points)
             
-                % Case A: uniformSurfaceOffset OFF
+                % Case A: Per-tile (mode = 1)
```
MyYo, line 347: remove mode =1 it's not relevant anymore

ThorlabsImager/test_yOCTPhotobleachTile_createPlan.m
```diff
@@ -344,38 +344,156 @@ function testUniformSurfaceOffset(tc)
...
-                    'Tiles should have different Z offsets when uniformSurfaceOffset = false');
-            
-                % Test with uniformSurfaceOffset, all z offsets must be equal
+
+                % All offsets must be equal
                 tc.verifyLessThan(range(dzB), 1e-6, ...
-                    'All Z offsets must match when uniformSurfaceOffset = true');
+                    'All Z offsets must match when surfaceCorrectionMode="origin-tile"');
+                
+                % Identify the centered tile to compare expected value
```
MyYo, line 375: origin tile

ThorlabsImager/test_yOCTPhotobleachTile_createPlan.m
```diff
@@ -344,38 +344,156 @@ function testUniformSurfaceOffset(tc)
...
+                % Case E: per-tile produces NaNs in a subset; uniform must lift them to zRef and move stage Z
+                % This checks that there are no errors bringing previous invalid tiles from failed focus validations
+
+                % Build a surface map that yields NaN only in two off center patches
+                S_subset = S;
+                patch1 = (X > 1.5 & X < 3.5 & Y > 1.5 & Y < 3.5);            % near (+2,+2)
+                patch2 = (X < -1.5 & X > -3.5 & Y < -1.5 & Y > -3.5);        % near (-2,-2)
+                S_subset.surfacePosition_mm(patch1 | patch2) = NaN;
+                
+                % First run per-tile (mode=1) to identify which tiles became NaN
```
MyYo, line 444: remove mode=1

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +119,182 @@
 assert(isscalar(json.nPasses), 'Only 1 nPasses is permitted for all lines');
 assert(isscalar(json.exposure), 'Only 1 exposure is permitted for all lines');
 
+% Normalize surfaceCorrectionMode (case/separator tolerant parsing)
+m = char(json.surfaceCorrectionMode);
```
MyYo, line 123: Why did you implement this way instead of 
switch(lower(json.surfaceCorrectionMode))

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -27,6 +27,10 @@
 %                                                   surfaceMap.surfacePosition_mm = surfacePosition_mm;
 %                                                   surfaceMap.surfaceX_mm        = surfaceX_mm;
 %                                                   surfaceMap.surfaceY_mm        = surfaceY_mm;
+%  surfaceCorrectionMode    "per-tile"      "per-tile":     In this mode, the algorithm computes and applies the best z offset
```
MyYo, line 30: Can also be "none" right?

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +119,182 @@
...
 % This function takes a photobleach plan, adjust it by Z according to the
-% surface map S
-function photobleachPlan = adjustPlanZ(S, photobleachPlan)
-    if isempty(S) % The case where tissue surface was not provided
+% surface map S and the requested surfaceCorrectionMode
+function photobleachPlan = applySurfaceCorrectionMode(photobleachPlan, json, v)
+% Make Z offset correction selection by mode.
+% Mode "per-tile" (default): Best z offset for each tile (lens field of view) 
+%                               individually.
+% Mode "origin-tile"       : Optimized Z offset for the origin tile
```
MyYo, line 185: Can also be none

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +119,182 @@
...
+            else
+                % Limit offset to 100 microns to prevent lens damage
+                zSurf_mm = max(min(zSurf_mm, 0.1), -0.1);
+                photobleachPlan(iXY).zOffsetDueToTissueSurface = zSurf_mm;
+            end
+        end
+
+        % Apply stage Z using the decided per-tile offsets
+        for iXY = 1:numel(photobleachPlan)
+            zOff = photobleachPlan(iXY).zOffsetDueToTissueSurface;
```
MyYo, line 233: merge with the loop above

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +119,182 @@
...
+% Mode "per-tile" (default): Best z offset for each tile (lens field of view) 
+%                               individually.
+% Mode "origin-tile"       : Optimized Z offset for the origin tile 
+%                               (the tile that is closest to x=0, y=0)
+
+    S = json.surfaceMap;
+    mode = json.surfaceCorrectionMode;
+
+    % If no surface map was provided, use zero offsets and return
+    if isempty(S)
```
MyYo, line 192: I recommend to refactor as following

if isempty(S)
   constantZCorrection_mm = 0;
   errIdConstantZCorrection ='';
elseif strcmp(S,'origin-tile')
   [constantZCorrection_mm, errId] = computeZCorrectionForTile(x_mm, y_mm, json.FOV, S, v);
else
   constantZCorrection_mm = []
   errIdConstantZCorrection =''
end

for iXY = 1:numel(photobleachPlan)
            x_mm = photobleachPlan(iXY).stageCenterX_mm;
            y_mm = photobleachPlan(iXY).stageCenterY_mm;
        
            % Compute Z offset for this tile
            if isempty(constantZCorrection_mm)
                [zSurf_mm, errId] = computeZCorrectionForTile(x_mm, y_mm, json.FOV, S, v);
           else 
                zSurf_mm = constantZCorrection_mm
                errId = errIdConstantZCorrection
end 

< REST OF THE CODE>


The reason I propose that because it standardize the treatment logic.
For example, rigtht now in the origin mode you forgot to Limit offset to 100 microns to prevent lens damage

MyYo (changes requested):
Thanks. See Refactor requests. Please make this highest priority.

ThorlabsImager/yOCTPhotobleachTile.m
alejoruge: Right, I think that's better and easier to read. I changed it now.

ThorlabsImager/yOCTPhotobleachTile.m
alejoruge: Got it. I think this is done now, please let me know.

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +121,163 @@
 assert(isscalar(json.nPasses), 'Only 1 nPasses is permitted for all lines');
 assert(isscalar(json.exposure), 'Only 1 exposure is permitted for all lines');
 
+% Normalize surfaceCorrectionMode (case/separator tolerant parsing)
```
MyYo, line 124: It's fine to not do the user's work for them.
If it's not 'per-tile','origin-tile' or 'none' just throw and error.

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +121,163 @@
...
+    S    = json.surfaceMap;
+    mode = json.surfaceCorrectionMode;
+
+    % Decide if we have a Z constant correction for all tiles
+    constantZCorrection_mm   = [];   % [] compute per-tile Z correction inside the loop
+    errIdConstantZCorrection = '';
+    idx0 = []; % for logging in origin-tile
+
+    % If no surface map OR mode 'none' was provided, use zero offsets and return
+    if isempty(S) || strcmp(mode,'none')
```
MyYo, line 198: Please assume mode is one of the three options
'per-tile','origin-tile' or 'none'.
All other options (including empty) should not be supported.

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +121,163 @@
...
+        % Make no adjustments: uniform zero correction
```
MyYo, line 199: please remove rows 200-202

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +121,163 @@
...
+%                               (the tile that is closest to x=0, y=0).
+% Mode "none"              : No surface-based Z correction (all offsets are zero).
+
+    S    = json.surfaceMap;
+    mode = json.surfaceCorrectionMode;
+
+    % Decide if we have a Z constant correction for all tiles
+    constantZCorrection_mm   = [];   % [] compute per-tile Z correction inside the loop
+    errIdConstantZCorrection = '';
+    idx0 = []; % for logging in origin-tile
```
MyYo, line 195: row 195 is not needed, I think you can remove

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +121,163 @@
...
+
+    % If no surface map OR mode 'none' was provided, use zero offsets and return
+    if isempty(S) || strcmp(mode,'none')
+        % Make no adjustments: uniform zero correction
         for iXY = 1:numel(photobleachPlan)
             photobleachPlan(iXY).zOffsetDueToTissueSurface = 0;
         end
+        constantZCorrection_mm   = 0;
+        errIdConstantZCorrection = '';
+        if v && strcmp(mode,'none')
```
MyYo, line 205: you can just if v, strcmp is redundant

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +121,163 @@
...
+                fprintf('%s applySurfaceCorrectionMode: origin zOffset is invalid (NaN). Using zero offset for all tiles.\n', datestr(datetime));
+            end
+            zRef    = 0;
+            errId0  = '';
+        end
+        constantZCorrection_mm  = zRef;   % clamp happens uniformly inside the loop
+        errIdConstantZCorrection = errId0;
+
+        % Set the uniform Z offset and photobleach all tiles
+        if isfield(photobleachPlan,'performTilePhotobleaching')
```
MyYo, line 234: I think you can remove 234-236

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +121,163 @@
...
+        [zRef, errId0] = computeZCorrectionForTile(x0_mm, y0_mm, json.FOV, S, v);
+        
+        % Fallback to zero if invalid (preserves previous behavior)
+        if isempty(zRef) || isnan(zRef)
+            if v
+                fprintf('%s applySurfaceCorrectionMode: origin zOffset is invalid (NaN). Using zero offset for all tiles.\n', datestr(datetime));
+            end
+            zRef    = 0;
+            errId0  = '';
+        end
```
MyYo, line 229: change to
        constantZCorrection_mm  = 0;  
        errIdConstantZCorrection = 0;
else
       constantZCorrection_mm  = zRef;   
        errIdConstantZCorrection = errId0;
end

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +121,163 @@
...
-        else
-            photobleachPlan(iXY).zOffsetDueToTissueSurface = NaN;
         end
     end
+
+    % Logging for origin-tile uniform value (after it has been clamped in the loop)
+    if strcmp(mode,'origin-tile') && v && ~isempty(idx0)
+        % Grab the (now clamped) value from any tile
+        zUni = photobleachPlan(1).zOffsetDueToTissueSurface;
+        fprintf('%s applySurfaceCorrectionMode: uniform Z offset = %.4f mm from centered tile %d applied to all tiles.\n', ...
```
MyYo, line 274: This printing doesn't happen for all other modes, I think you can skip it.
Remove 271-276

MyYo (changes requested):
Thanks, additional small refactoring is needed

ThorlabsImager/yOCTPhotobleachTile.m
alejoruge: Ok I removed the variations, please let me know if it looks good now

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +121,146 @@
...
+switch regexprep(lower(strtrim(char(json.surfaceCorrectionMode))),'[\s_–—−-]+','-') % Map spaces/underscores/dashes to a single '-'
+    % Canonicalize to either 'per-tile', 'origin-tile' or 'none'
+    case {'per-tile'}
+        json.surfaceCorrectionMode = 'per-tile';
+    case {'origin-tile'}
+        json.surfaceCorrectionMode = 'origin-tile';
+    case {'none'}
+        json.surfaceCorrectionMode = 'none';
+    otherwise
+        error('surfaceCorrectionMode must be "per-tile" (default), "origin-tile", or "none".');
```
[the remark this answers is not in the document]
alejoruge: Okay! Got it, resolved now

ThorlabsImager/yOCTPhotobleachTile.m
alejoruge: I think one example that would break would be  the **script_PhotobleachXYPattern** because it doesn't send the surfaceMap

MyYo: I see so the issue is if surfaceMap is empty and correction is not none.
This is an input check and should be at the input check section above

ThorlabsImager/yOCTPhotobleachTile.m
```diff
@@ -114,77 +121,145 @@
...
+    case {'per-tile'}
+        json.surfaceCorrectionMode = 'per-tile';
+    case {'origin-tile'}
+        json.surfaceCorrectionMode = 'origin-tile';
+    case {'none'}
+        json.surfaceCorrectionMode = 'none';
+    otherwise
+        error('Invalid surfaceCorrectionMode value: %s', surfaceCorrectionMode);
+end
+
```
MyYo, line 135: Add here 
if isempty(surfaceMap) && ~strcmp(json.SurfaceCMode,'none')
error('Surface map wasn't provided, surfaceCorrectionMode cannot be %s,'surfaceCorrectionMode)

MyYo (approved):
One small correction needed.
You can push it and merge.

alejoruge:
Please help me with thist last review. I am having three options here to address the last feedback properly:


Problem: When surfaceMap is not passed (default = empty) and/or surfaceCorrectionMode is not passed (default = "per-tile"), the defaults conflict and trigger the error. The problem is these are the defaults, meaning all functions call in the repositories currently trigger the error.


In order to address this I have these three options:

1. What I just committed: This follows the logic you requested, but checks whether the user passed the mode, or if it was taken by the default. Depending on this, it will throw the error (if user screwed it up) or set the mode to none instead (because this wasn't requested on purpose anyways).

2. We will need to modify all calls in all repositories that call yOCTPhotobleachTile, including some tests, because otherwise the function will fail when we merge because they are not currently passing "none" mode.

3.  We can set the "none" mode as default so that it won't reach the error when nothing is passed (default empty surfaceMap + default "none" mode won't trigger the error).

I also added a tester, will modify if you consider option 2 or 3 are better.

Please let me know what you think and I'll do it right away! Thank you

MyYo (approved):
Good point. In that case just set default to 'none'

PR #67, opened 2025-08-22
Use update in yOCTScanTile rename

alejoruge:
Previous update in yOCTScanTile renamed nPixels to nPixelsInEachTile

Here we add this new names in places where it was creating errors during scanning (in yOCTScanTile) and reconstruction (in yOCTProcessTiledScan_createDimStructure)


MyYo (approved):
Great looks good

PR #72, opened 2025-08-28
Fix FactorX for Y-axis stitching issue for 10x OCTP900 probe

alejoruge:
Set FactorX positive in Olympus10xW OCTP900 to match physical Y direction and prevent row inversion during stitching with 10x.


alejoruge:
screenshot

MyYo (approved):

PR #75, opened 2025-10-14
Add skipHardware early return to yOCTScanGlassSlideToFindFocusAndDisp…

alejoruge:
…ersionQuadraticTerm

Add skipHardware early return to glass slide focus/dispersion function

Current situation & Problem

- yOCTScanGlassSlideToFindFocusAndDispersionQuadraticTerm accepts skipHardware parameter but doesn't handle it
- Function attempts to read non-existent scan files when skipHardware=true
- Causes errors in simulation mode after yOCTScanTile returns early
- No way to skip focus/dispersion measurement during testings

Release Notes

- Added early return when skipHardware=true (lines 41-51)
- Returns NaN for both dispersionQuadraticTerm and focusPositionInImageZpix
- NaN values clearly indicate unmeasured parameters in simulation mode
- Prints informative message in verbose mode
- Prevents file read errors and unnecessary processing
- Consistent with skipHardware behavior in yOCTScanTile and other myOCT functions


MyYo (changes requested):
There is implementation of skipHardware in line 92.
Can you explain more about why you implement here?

My hope is that by using skipHardware, we can still debug the function though just not the hardware part. So if there is a way not to bypass it altogether I think its better.

Can you explain the scenario in which you experienced an issue?

alejoruge:
Oh I see, I didn't think about it that way. The problem comes at line 88 where it tries to read ScanInfo.json, but that file doesn't exist when skipHardware=true because yOCTScanTile doesn't create it.

I was thinking about an alternative: what if we check skipHardware right before line 88 and generate mock data only for that part? Would you like me to try that?

Right now, every time we have skipHardware = true, we need to comment the whole function out because it runs the error..

Applications/yOCTScanGlassSlideToFindFocusAndDispersionQuadraticTerm.m
```diff
@@ -73,24 +83,44 @@
...
+                'v',in.v  ...
+                );
+        end
+        
         atFocusIndex = NaN;
         atFocusIntensity = 0;
         for scanI = 1:length(json.octFolders)
-            [interf, dim] = yOCTLoadInterfFromFile(...
-                [tempFolder json.octFolders{scanI}], 'YFramesToProcess',1);
+            % For simulated data, pass OCTSystem explicitly to avoid detection error
```
MyYo: No need for duplication here you can just use 
'''
[interf, dim] = yOCTLoadInterfFromFile([tempFolder json.octFolders{scanI}], ...
                    'YFramesToProcess', 1, ...
                    'OCTSystem', json.OCTSystem);
'''
For both cases

Applications/yOCTScanGlassSlideToFindFocusAndDispersionQuadraticTerm.m
```diff
@@ -107,6 +137,13 @@
             end
         end
 
+        % In simulation mode, intensity scores may be uniform across all scans
```
MyYo: I don't think this is needed. If all intensities have same value a valid atFocusIndex is still selected.

MyYo (changes requested):

MyYo (approved):
looks good.
Small comment

PR #82, opened 2025-10-31
Create a unified framework for loading hardware library to memory and initializing OCT scanner

MyYo:
Using persistent control pattern was recommended by ChatGPT.

Prompt: 
"In matlab, I need to load a class to memory which is used by multiple functions. It's cumbersome to pass the library over and over and also declare it in every sub call. Is there a standard way to solve this?"


MyYo:
@alejoruge please review this pull request, approve and merge. Don't add additional functionality to this pull request.

alejoruge (approved):
I think this is a brilliant idea. Will merge and work from this in the PRs

PR #85, opened 2025-11-05
Add verbose mode parameter to yOCTLoadHardwareLib

alejoruge:
- Add 'v' parameter to yOCTLoadHardwareLib (default: false)
- ⁠Pass chosen verbose mode to yOCTImportPythonModule
- ⁠Add octSystem variable to all demos as user input


ThorlabsImager/yOCTLoadHardwareLib.m
```diff
@@ -32,10 +33,14 @@
     skipHardware = false;
 end
 
-validSystems = {'GAN632', 'Ganymede'};
+if ~exist('v','var')
```
MyYo: Move this to the very top, row 14.

PR #87, opened 2025-11-05
Refactor yOCTStageInit to support multiple OCT systems

alejoruge:
Changes:

-  ⁠Added call to yOCTLoadHardwareLib() to detect OCT system type -⁠  ⁠Implemented switch statement for 'ganymede' (C# DLL) and 'gan632' (Python) systems -⁠  ⁠Ganymede case: Uses existing ThorlabsImagerNET.ThorlabsImager.yOCTStageInit() -⁠  ⁠Gan632 case: Uses Python module with octSystemModule.yOCTStageInit_1axis() -⁠  ⁠Added verbose mode messages for both systems to indicate which control method is active -⁠  ⁠Updated Motion Range Test section with system-specific implementations -⁠  ⁠Added skipHardware mode support to return (0,0,0) for simulation scenarios

This change maintains backward compatibility while enabling support for the new Gan632 system.

### DIFF
--- a/ThorlabsImager/yOCTStageInit.m
+++ b/ThorlabsImager/yOCTStageInit.m
@@ -127,7 +127,7 @@ if ~skipHardware
                         gStageCurrentStagePosition_StageCoordinates(i));
                     
                 case 'gan632'
-                    % GAN632: Use Python module for motion test
+                    % Gan632: Use Python module for motion test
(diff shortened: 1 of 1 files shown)


MyYo (changes requested):
Looks good. Minor change needed.

MyYo (approved):
Looks good, thanks

PR #90, opened 2025-11-06
Handle absolute repo paths and include .py packages in path detection

alejoruge:
♻️ Current Situation & Problem
The function assumed all repoName inputs were relative paths and only detected packages implemented as folders.
This caused incorrect path resolutions when using absolute paths and missed Python packages distributed as single .py files.

⚙️ Release Notes:
	•	✅ Added logic to handle both absolute and relative repoName inputs.
	•	🧩 Updated repoBase assignment to check with isfolder() before appending pwd.
	•	📦 Extended package detection to include .py files alongside folders.
	•	📝 Improved comments and readability of path-handling logic.

📚 Documentation
Inline comments added to clarify absolute vs relative path handling and .py detection behavior.

✅ Testing
	•	Verified correct handling for both relative and absolute repo paths.
	•	Confirmed .py package files are correctly added to sys.path.
	•	No regression found in previous path detection behavior.

This updates adds compatibility with current new OCT system setup.

### DIFF
--- a/Utils/yOCTImportPythonModule.m
+++ b/Utils/yOCTImportPythonModule.m
@@ -80,7 +80,14 @@ pyexe = char(pe.Executable);
 if isempty(in.repoName)
     repoBase = pwd;
 else
-    repoBase = fullfile(pwd, char(in.repoName));
+    % Check if repoName is already an absolute path
+    if isfolder(char(in.repoName))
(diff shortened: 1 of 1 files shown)


MyYo (approved):
Ideally you would have a tester. But I will approve it as it's a small change.

PR #98, opened 2025-12-06
Update stage functions and set velocity

albertchen1375:
### :recycle: Current situation & Problem

- Stage velocity and acceleration were at 1 mm/s and 1 mm/s^2
- This was comparably slow and left room for improvement

### :gear: Release Notes

- Stages are now initialized with a velocity of 2 mm/s and acceleration of 3 mm/s^2
- Created a function to set velocity and acceleration of stage
- Created a function to get velocity and acceleration of stage

### :books: Documentation

- Updated the docstring of stage function "yOCTStageInit_1axis"

### :white_check_mark: Testing

- Tested with Demo_ScanAndProcess3D.m repetitively
- Independent scripts were tested with testing scripts that are not being tracked by Github (can improve in further PR)

### DIFF
--- a/ThorlabsImager/ThorlabsImagerPython/thorlabs_imager_stage.py
+++ b/ThorlabsImager/ThorlabsImagerPython/thorlabs_imager_stage.py
@@ -29,13 +29,13 @@ _stage_serial_numbers = {
 _stage_handles = {}
 
 
-def yOCTStageInit_1axis(axes: str) -> float:
+def yOCTStageInit_1axis(axes: str, max_velocity_mm_sec: float = 2.0, max_acceleration_mm_s_2: float = 3.0) -> float:
     """Initialize stage for one axis and return current position in mm.
-    
-    Automatically sets velocity to 2.0 mm/s and acceleration to 3.0 mm/s².
     
     Args:
         axes (str): Axis identifier ('x', 'y', or 'z')
+        max_velocity_mm_sec (float): Maximum velocity in mm/s (default: 2.0)
+        max_acceleration_mm_s_2 (float): Maximum acceleration in mm/s² (default: 3.0)
     
     Returns:
         float: Current position in mm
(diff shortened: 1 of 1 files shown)


ThorlabsImager/ThorlabsImagerPython/thorlabs_imager_stage.py
```diff
@@ -30,7 +30,20 @@
 
 
 def yOCTStageInit_1axis(axes: str) -> float:
-    """Initialize stage for one axis and return current position in mm."""
+    """Initialize stage for one axis and return current position in mm.
```
MyYo: Better to give the option to the user, but set defaults.
yOCTStageInit_1axis(axes: str, max_velocity_mm_sec=2, max_acceleration_mm_s_2=3)

ThorlabsImager/ThorlabsImagerPython/thorlabs_imager_stage.py
```diff
@@ -158,9 +184,101 @@ def yOCTCloseAllStages():
     gc.collect()
 
 
+def yOCTStageSetVelocity_1axis(axis: str, max_velocity_mm_s: float, acceleration_mm_s2: float) -> None:
```
MyYo: I think we should set speed only on initialization unless you see a use case for this right now

ThorlabsImager/ThorlabsImagerPython/thorlabs_imager_stage.py
```diff
@@ -158,9 +184,101 @@ def yOCTCloseAllStages():
...
+            vel_conv  # max_velocity (already in device units as int)
+        )
+        
+    except XADeviceException as e:
+        raise RuntimeError(f"XADeviceException setting velocity: {e.error_code}")
+    except Exception as e:
+        raise RuntimeError(f"Error setting velocity for axis {axis}: {e}")
+
+
+def yOCTStageGetVelocity_1axis(axis: str) -> tuple:
```
MyYo: Function naming doesn't follow function's actions.
It doesn't give the current velocity of the stage, it's returning the maximum velocity and acceleration.

Unless there is a use case for that function right now, you can also remove.

PR #105, opened 2025-12-17
Unzip .oct files during PROCESSING

alejoruge:
**Problem:**
Processing scans created with unzipOCTFile=false during scanning fails when reconstructing them:

Error using imageDatastore (line 139)
Cannot find files or folders matching: '.../Data01/Header.xml'

The processing code expects Header.xml to exist, but when scans are saved compressed (.oct files), it doesn't check if decompression is needed before attempting to read.

**Solution:**

Created **yOCTUnzipTiledScan** to provide explicit user control over decompression:
- Reads 'ScanInfo.json' to find all OCT folders in a tiled scan
- Decompresses '.oct`'files in parallel using 'yOCTUnzipOCTFolder'
- Reports success/failure status for each folder
- Only deletes source '.oct' files after verifying successful extraction

Added call to 'yOCTProcessTiledScan':
- New parameter '**unzipBeforeProcessing**' (default: 'false')
- User explicitly chooses when to decompress, avoiding automatic behavior

Improvements in 'yOCTUnzipOCTFolder':
- Automatically detects and fixes malformed filenames from 7-Zip extraction (this causes severe error when decompressing damaging entire scans)
- Verifies 'Header.xml' exists before deleting source '.oct'
- Preserves failed extractions for debugging

### DIFF
--- a/Processing/yOCTProcessTiledScan.m
+++ b/Processing/yOCTProcessTiledScan.m
@@ -111,8 +111,6 @@ end
 json = awsReadJSON([tiledScanInputFolder 'ScanInfo.json']);
 
 %% Unzip compressed .oct files if they exist in the data folder
-% When scans were saved with unzipOCTFile=false, this step decompresses
-% all .oct files before parallel processing to avoid errors during processing
 unzipResults = yOCTUnzipTiledScan(tiledScanInputFolder, ...
     'deleteCompressedAfterUnzip', true, ...
     'v', v);


MyYo (changes requested):
Lets talk about this. 
There is an architecture issues with this solution

alejoruge:
New approach to the problem presented here now

LoadSave/yOCTUnzipTiledScan.m
```diff
@@ -0,0 +1,177 @@
...
+%   5. Only deletes compressed files that were successfully unzipped
+%
+% INPUTS:
+%   tiledScanInputFolder         - Path to tiled scan folder containing ScanInfo.json
+%   'deleteCompressedAfterUnzip' - Delete .oct files after successful unzip (default: true)
+%   'v'                          - Verbose mode with detailed progress reporting (default: false)
+%
+% OUTPUTS:
+%   results - Structure with fields:
+%       .alreadyUnzipped      - Folder names already unzipped from previous runs
```
MyYo: Folder names already unzipped from previous runs --> Folder names already unzipped

LoadSave/yOCTUnzipTiledScan.m
```diff
@@ -0,0 +1,177 @@
...
+v = in.v;
+
+% Fix input path
+if ~strcmp(in.tiledScanInputFolder(end), '/') && ~strcmp(in.tiledScanInputFolder(end), '\')
+    tiledScanInputFolder = awsModifyPathForCompetability([in.tiledScanInputFolder '/']);
+else
+    tiledScanInputFolder = awsModifyPathForCompetability(in.tiledScanInputFolder);
+end
+
+%% Load ScanInfo.json to get list of OCT folders
```
MyYo: Question about implementation.
Why not instead of all this logic below, use awsls(tiledScanInputFolder) and unzip any file with the name .oct?

LoadSave/yOCTUnzipOCTFolder.m
```diff
@@ -57,6 +57,23 @@ function yOCTUnzipOCTFolder(OCTFolderZipFileIn,OCTFolderOut,isDeleteOCTZippedFil
 
 %% Preform Unzip
 
+% Clean destination directory content if it exists (keeping the .oct file)
+if exist(OCTUnzipToDirectory,'dir')
+    % Only delete data folder and extracted files, not the .oct itself
+    dataFolderToClean = fullfile(OCTUnzipToDirectory, 'data');
+    if exist(dataFolderToClean, 'dir')
+        rmdir(dataFolderToClean, 's');
```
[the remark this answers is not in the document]
alejoruge: Let me try to explain why I believe this wouldn't be a good idea:

This block handles re decompression after failed extractions (PC restart, power loss, manual cancellation by mistake, etc.). So decompression was interrupted, this block helps.

How it works:

**Header.xml** is the last file extracted -> if it exists = successful decompression for that .OCT

**Compressed .oct** is only deleted after confirming Header.xml exists, as a last step

If decompression failed, data/ folder exists but incomplete (no **Header.xml**)

This block detects incomplete extraction and automatically re decompresses this file.

Why throwing an error is problematic:

Since yOCTUnzipTiledScan uses parfor to decompress multiple folders in parallel, an interruption affects many folders simultaneously

If we throw an error:

- The user must manually fix each incomplete folder (doesn't know how, and error only shows 1)
- Error repeats for every failed folder in the parfor loop after each manual fix
- User could accidentally damage data during manual cleanup because they don't know what's going on
- Process could take hours of tedious manual work


This block currently:

- Automatically cleans incomplete extractions and re decompresses if header file is missing (incomplete/corrupted)
- Handles all failed folders in one execution
- No user intervention needed
- Recovery from interruptions is clean, a ccurate and quick

This is the reason I recommend keeping this block to handle batch re decompression gracefully.. Please let me know if you think otherwise.

LoadSave/yOCTUnzipTiledScan.m
```diff
@@ -0,0 +1,177 @@
...
+addRequired(p, 'tiledScanInputFolder', @ischar);
+addParameter(p, 'deleteCompressedAfterUnzip', true, @islogical);
+addParameter(p, 'v', false, @islogical);
+parse(p, tiledScanInputFolder, varargin{:});
+
+in = p.Results;
+v = in.v;
+
+% Fix input path
+if ~strcmp(in.tiledScanInputFolder(end), '/') && ~strcmp(in.tiledScanInputFolder(end), '\')
```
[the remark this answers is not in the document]
alejoruge: Much better as well, thanks. Resolved!

LoadSave/yOCTUnzipTiledScan.m
alejoruge: Done!

LoadSave/yOCTUnzipTiledScan.m
alejoruge: Good question. I believe you're suggesting changing:

json = awsReadJSON(scanInfoPath);
folders = json.octFolders; // ['Data01', 'Data02', ...]

to:

folders = awsls(tiledScanInputFolder); 
folders = folders(~strcmp(folders, 'ScanInfo.json')); // filter files

**If we keep logic but just change folder discovery:**

**JSON approach:** Processes the scan's official folder list (defined at scan time)
**awsls approach:** Processes whatever folders currently exist

My concern: Since function is yOCTUnzipTiledScan, I think it should unzip **the scan** (as defined by metadata), not "whatever happens to be there". If someone adds 'DataBackup01/' manually, awsls would try to unzip it.

**Also:** Reading ScanInfo.json validates the scan happened correctly.

Both would work technically. I just think JSON better matches the function's intent. Happy to change if you disagree.

In regards to logic, if we use one that checks everything and just unzip any .oct file without the rest of the logic, we would lose:**

1. **Header.xml check**: we currently skips folders already unzipped. Without this, re-running would re-decompress everything (very slow)
2. **Parallel processing**: parfor decompresses several folders simultaneously. Sequential would be ~10x slower
3. **Error handling** : our try-catch allows continuing if one folder fails (like if one corrupted file exists, it stops instead of continuing with the rest of them and just report the one that failed due to corruption during scan/saving).
4. **Post-unzip validation**: This verifies Header.xml exists after decompression to confirm success, if it doesn't it failed during decompression (maybe the computer failed after hours of decompression, or user by mistake stopped)
5. **Progress tracking**: this reports "10 already done, 38 newly unzipped, 2 failed (Data05, Data23)", we would just lose this
6. **Preserve .oct on failure**: this keeps compressed file for easy recognition for debugging if decompression fails

Please let me know what you think

PR #106, opened 2025-12-17
Prevent redundant hardware initialization

alejoruge:
**Problem**
Redundant initialization attempts: **yOCTStageInit** and **yOCTTakeImagesTile** were calling ThorlabsImagerNETLoadLib(), attempting to reload the .NET assembly that was already loaded by **yOCTLoadHardwareLib** at the start of the scan session and this caused errors.

**Solution**
Removed redundant library loads: Deleted ThorlabsImagerNETLoadLib() calls from yOCTStageInit and yOCTTakeImagesTile since the hardware library is already initialized once at the beginning via yOCTLoadHardwareLib().

### DIFF
--- a/ThorlabsImager/yOCTLoadHardwareLib.m
+++ b/ThorlabsImager/yOCTLoadHardwareLib.m
@@ -72,15 +72,25 @@ octSystemName = lower(octSystemName);
 switch(octSystemName)
     case 'ganymede'
         % Ganymede: C# library 
-
-        % Verify that library wasn't loaded before
-        if ~isempty(which('ThorlabsImagerNET.ThorlabsImager')) 
-            error('ThorlabsImagerNET loaded before, this should never happen.');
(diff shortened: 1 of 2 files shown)


MyYo (changes requested):
PR covers two ideas. Please split.

ThorlabsImager/yOCTStageInit.m
```diff
@@ -57,7 +57,7 @@
             if (v)
                 fprintf('%s [Ganymede] Initializing C# DLL-based stage control (3 axes)...\n', datestr(datetime));
             end
-            ThorlabsImagerNETLoadLib();
+            % Library already loaded by yOCTLoadHardwareLib()
```
MyYo: I prefer you delete the line rather than comment it out

MyYo (changes requested):

PR #107, opened 2025-12-18
Fix floating point precision issues

alejoruge:
**Capability:** Can the system handle scan ranges that are valid multipliers but fail due to floating-point precision?

♻️ **Current situation & Problem**
When scanning with certain valid tile configurations, the system would throw an error due to floating-point precision issues in multiplier validation. For example:

yOverall_mm         = [-4.42 4.42]
octProbeFOV_mm = 0.34 mm

Error message: "User requested to scan y=[-4.42 to 4.42], but the size of the scan is not a multiplier of the FOV: 0.34"

The issue occurs very often during large scans. It happens because 8.84 / 0.34 = 26.0000000000001 in binary floating-point arithmetic, causing the exact equality check round(x) ~= x to fail even though mathematically it should be 26.0 exactly.

⚙️ **Release Notes**

Fixed floating-point precision issue in **yOCTScanTile_XYRangeToCenters**.m by using tolerance based comparison (tol = 1e-6 mm) instead of exact equality
The tolerance of 1 nanometer is physically insignificant compared to the system's precision in microns and ensures valid scan configurations are accepted

📚 Documentation
Added inline comment explaining the tolerance based approach to avoid floating-point precision issues

✅ Testing

-Tested with the problematic configurations that previously failed
-Verified that tile center calculations remain accurate and valid with the tolerance adjustment

### DIFF
--- a/ThorlabsImager/yOCTScanTile_XYRangeToCenters.m
+++ b/ThorlabsImager/yOCTScanTile_XYRangeToCenters.m
@@ -26,14 +26,14 @@ xSize_mm = diff(xRange_mm);
 ySize_mm = diff(yRange_mm);
 
 % Use tolerance for floating point comparison (avoid precision issues)
-tol = 1e-6;
+tolerance_mm = 1e-6;
 
 if ((xSize_mm > octProbeFOV_mm) && ...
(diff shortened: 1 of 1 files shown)


MyYo (approved):
Thanks,
Good catch - please make minor change and merge.

PR #108, opened 2025-12-18
Fix scan function signature

alejoruge:
yOCTScanTile lost it's signature during previous commits and needs it in order to work without errors


alejoruge:
**Problem**
**Missing function signature:** yOCTScanTile.m was missing its function declaration line from previous merges, causing MATLAB to fail when calling it from other scripts.

**Solution**
**Restored function signature**: Added back the missing function declaration line to yOCTScanTile.m

PR #131, opened 2026-03-31
Fold initStage into yOCTHardware - eliminate stage globals

alejoruge:
This is the first proposal. Please consider the tradeoffs of the approach:

**PROBLEM**

- yOCTHardware_initStage stores stage state in 3 global variables (goct2stageXYAngleDeg, gStageCurrentStagePosition_OCTCoordinates, gStageCurrentStagePosition_StageCoordinates) that live outside the persistent struct gOCTHardwareStatus. 

- When yOCTHardware('teardown') runs, it clears the struct but the globals remain, leaving stale stage state in memory. yOCTStageMoveTo also reads/writes those globals directly, bypassing the lifecycle manager.

**SOLUTION**

- Fold yOCTHardware_initStage into yOCTHardware as an 'initStage' command. 

- Add '**stageStatus**' (**read** only query) and '**updateStagePosition**' (**write**) commands so all stage state lives in gOCTHardwareStatus. 

- Update all callers (yOCTScanTile, yOCTPhotobleachTile, yOCTTakeImagesTile, yOCTStageMoveTo, yOCTTissueSurfaceAutofocus). 

- Delete yOCTHardware_initStage.m.

Tradeoffs with this approach:

- The 4 output variables of yOCTHardware change meaning per command (e.g. out1 is octSystemModule for 'status' but x0 for 'initStage', out1 is angle for 'stageStatus'). Callers need to know which command was used to interpret the outputs correctly.

- inputParser accepts all parameters regardless of command, passing 'posOCT' to 'init' is silently ignored instead of erroring.

- initScanner and closeScanner remain as separate helper files while initStage is now inline, creating an asymmetry.


Applications/yOCTTissueSurfaceAutofocus.m
```diff
@@ -171,7 +171,8 @@
     
     % Move the stage to bring tissue into focus if needed
     if needMove
-        [~,~,z0] = yOCTHardware_initStage();  % query current Z
+        [~, posOCT, ~] = yOCTHardware('stageStatus');
```
MyYo: stageStatus doesn't indicate an action, maybe getStageStatus?

ThorlabsImager/test_yOCTHardware.m
```diff
@@ -130,5 +130,115 @@ function testNoCommandErrors(testCase)
                 'myOCT:yOCTHardware:noCommand');
         end
 
+        %% initStage
+
+        %% initStage returns (0,0,0) when skipHardware is true
+        function testInitStageReturnsOrigin(testCase)
+            yOCTHardware('init', 'OCTSystem', 'Gan632', 'skipHardware', true);
+            [x0, y0, z0] = yOCTHardware('initStage');
```
MyYo: What is the use case for initiStage? Shouldn't this be part of init hardware?

ThorlabsImager/yOCTHardware.m
```diff
@@ -31,8 +31,21 @@
 %   'reset'      - Clear all persistent variables without closing hardware.
 %                  yOCTHardware('reset')
 %
+%   'initStage'  - Initialize translation stage, return position.
```
MyYo: I'm not sure what is the use case to have separate initStage which is distinctive from init.

ThorlabsImager/yOCTHardware.m
```diff
@@ -31,8 +31,21 @@
...
+%                  [x0, y0, z0] = yOCTHardware('initStage', ...)
+%                  Optional: 'oct2stageXYAngleDeg', deg, 'minPosition', vec,
+%                            'maxPosition', vec, 'v', tf
+%                  Requires init to be called first.
+%
+%   'stageStatus' - Return stage state (angle, OCT position, stage position).
+%                  [angle, posOCT, posStage] = yOCTHardware('stageStatus')
+%                  Errors if initStage was never called.
+%
+%   'updateStagePosition' - Update tracked stage position after movement.
```
MyYo: update is a bit confusing, do you mean move?

ThorlabsImager/yOCTHardware.m
```diff
@@ -31,8 +31,21 @@
...
+%                  yOCTHardware('updateStagePosition', 'posOCT', vec3, 'posStage', vec3)
```
MyYo: Explain what 'posOCT' and 'posStage' mean

MyYo (changes requested):

ThorlabsImager/yOCTHardware.m
alejoruge: Got it, used only for move and internally handle logic for angle

MyYo (changes requested):
Sorry to go back and forth on this, but the design pattern is still a little off.

For OCT the design pattern is
1. yOCTHardware initializes and sets the OCT hardware
2. yOCTScanTile, ThorlabsImagerNET.ThorlabsImager.yOCTPhotobleachLine etc preform the OCT actions like scan photobleach etc.

In current design pattern yOCTHardware does both stage hardware initialization as well as stage actions (yOCTHardware('moveStage').

If possible I prefer a design pattern that 
1. Uses yOCTHardware to initialize stage
2. Uses existing functions to perform stage actions (move, get position) etc

Applications/yOCTTissueSurfaceAutofocus.m
```diff
@@ -171,7 +171,7 @@
     
     % Move the stage to bring tissue into focus if needed
     if needMove
-        [~,~,z0] = yOCTHardware_initStage();  % query current Z
```
MyYo, line 174: Bring back the comment

ThorlabsImager/yOCTHardware.m
```diff
@@ -32,7 +35,7 @@
 %                  yOCTHardware('reset')
 %
 % VALID COMMAND SEQUENCES:
-%   Most common:         init -> status -> teardown
```
MyYo, line 35: I think init ->status->teardown is still a more common for scanning

ThorlabsImager/yOCTHardware.m
```diff
@@ -9,10 +9,13 @@
 %   scannerInitialized - True when scanner is currently initialized
 %
 % COMMANDS:
-%   'init'       - Load module and initialize scanner.
+%   'init'       - Load module, initialize scanner, and initialize translation stage.
 %                  yOCTHardware('init', 'OCTSystem', name, 'skipHardware', tf, ...
-%                               'octProbePath', path, 'v', tf)
+%                               'octProbePath', path, 'oct2stageXYAngleDeg', deg, ...
```
MyYo: I think you need to explan the meaning of each of the parameters

ThorlabsImager/yOCTHardware.m
```diff
@@ -32,7 +35,7 @@
 %                  yOCTHardware('reset')
 %
 % VALID COMMAND SEQUENCES:
-%   Most common:         init -> status -> teardown
+%   Most common:         init -> yOCTGetStagePosition -> yOCTStageMoveTo -> teardown
```
MyYo: Another comment, how is VerifyInit different from 'status' - should we just use 'status'

MyYo (changes requested):
Looks good, a few minor comments remaining

alejoruge:
This is the new changes in this refactor:

- One init, one job. yOCTHardware('init', ...) now fully initializes OCT + stage in a single call. It no longer accepts minPosition/maxPosition (previously used for motion test) because demos and scripts DO NOT have them yet, and the stage rotation angle is read automatically from the probe INI when octProbePath is provided, so demos and scripts don't need to change.

- Motion range test is its own function due to the previous point. The [**minPosition, maxPosition**] travel test moved out of init into a new helper **yOCTVerifyMotionRange**(minPos, maxPos, v). It can be called standalone after init, and it also registers the verified range in globals.

- **yOCTStageMoveTo** enforces the registered range. If **yOCTVerifyMotionRange** has been called, **yOCTStageMoveTo** errors on any move that falls outside the verified window. If the range was never registered, behavior is unchanged (backward compatible).

- **verifyInit** is gone. status now performs the same checks (errors when uninitialized, or when the scanner is required but not opened). Fewer commands, same guarantees.

- Callers cleaned up. yOCTScanTile, yOCTPhotobleachTile, yOCTScan3DVolume, yOCTTakeImagesTile, yOCTTissueSurfaceAutofocus, yOCTSimulateTileScan now use a single init upstream and status / yOCTVerifyMotionRange downstream: no more two-phase init.

ThorlabsImager/yOCTHardware.m
alejoruge: It was only rejecting the initialization with error, but it was being used each time as a redundant because it was right before status. I changed it to make it simpler as I agree. They now are one. status is the only guard.

ThorlabsImager/yOCTHardware.m
alejoruge: Got it, done. I also refactored the logic. The problem was, we were using max and min positions to test motion of the stage. Now, yOCTVerifyMotionRange will do that in the functions whenever needed, but stage will be initialized.

ThorlabsImager/yOCTHardware.m
alejoruge: Got it, sounds good to me. Reverted most common comment.

Applications/yOCTTissueSurfaceAutofocus.m
alejoruge: Got it, resolved.

ThorlabsImager/yOCTHardware.m
```diff
@@ -9,19 +9,19 @@
...
-%                  yOCTHardware('init', 'OCTSystem', name, 'skipHardware', tf, ...
-%                               'octProbePath', path, 'v', tf)
-%                  octProbePath can be '' when skipHardware=true.
+%   'init'       - Load hardware module, initialize scanner, and initialize
+%                  translation stage.
+%                  yOCTHardware('init', 'OCTSystem', name, ...)
+%                  Parameters:
+%                    OCTSystem            - System name: 'Ganymede' or 'Gan632'. Required.
+%                    octProbePath         - Path to the probe .ini file. Required unless skipHardware=true.
+%                    oct2stageXYAngleDeg  - (default NaN) Override for the stage rotation angle in degrees.
```
MyYo: I would remove this. If override is needed, rewrite the probe.

ThorlabsImager/yOCTStageMoveTo.m
MyYo, line 1: Can you verify what was changed. It seems that you deleted the entire function just to paste it again.
It's not as safe from testing perspective, better to do a minimal change needed

ThorlabsImager/yOCTTakeImagesTile.m
```diff
@@ -82,7 +81,8 @@
     fprintf('%s Initialzing Hardware...\n\t(if Matlab is taking more than 2 minutes to finish this step, restart hardware and try again)\n',datestr(datetime));
 end
 
-[x0,y0] = yOCTHardware_initStage(in.oct2stageXYAngleDeg);
+yOCTHardware('init', 'oct2stageXYAngleDeg', in.oct2stageXYAngleDeg);
```
MyYo: I suggest remove support for oct2stageXYAngleDeg - just get it from OCT probe file

ThorlabsImager/yOCTVerifyMotionRange.m
```diff
@@ -0,0 +1,92 @@
+function yOCTVerifyMotionRange(minPosition, maxPosition, v)
+% Verify the stage can reach the requested motion range without collisions.
+% Moves each axis to its minimum, then maximum, then back to the origin.
```
MyYo, line 3: Just clarify if minPosition, maxPosition is in OCT coordinate system or in stage coordinate system?
There is a slight difference due to angle rotation.

I'm not sure which one is better... what do you think?

MyYo (changes requested):

ThorlabsImager/yOCTVerifyMotionRange.m
alejoruge: Oh you're right.. I think that since what we actually want to verify is that the stage can reach every point it will visit during the scan, the motion test should apply the same rotation...

The motion test has always been using **OCT coordinate** values directly as stage offsets, no angle rotation applied. So technically it was testing the wrong positions. (never noticed as it was a really small difference).

Meanwhile, actual moves (**yOCTStageMoveTo**) correctly apply the rotation, so all scans and photobleach always went to the right place.

We can fix it now, it's a small change in yOCTVerifyMotionRange, and it would make the test geometrically consistent with the actual scan motion for the first time. 

I will apply this as I think it's the right approach.

ThorlabsImager/yOCTTakeImagesTile.m
alejoruge: Got it, done

ThorlabsImager/yOCTVerifyMotionRange.m
MyYo: Just to clarify

NO ANGLE CORRECTION = Stage Coordinates 
WITH ANGLE CORRECTION = OCT Coordinate

I think there is a mistake in your note.

alejoruge: Right! Sorry, the motion test was taking values in OCT coordinates (from the scan grid) but passing them directly to the motor without applying the angle rotation, so it was effectively treating them as Stage coordinates (which was wrong). 

This is now fixed: **yOCTVerifyMotionRange** applies the same **Oct2StageXYAngleDeg** rotation that **yOCTStageMoveTo** has always used, so the motion test now visits the geometrically correct positions. 

Sorry for the confusion in my earlier note.

ThorlabsImager/yOCTStageMoveTo.m
alejoruge: Right, this is fixed now. Sorry about the inconvenience but you should be able to see the actual changes now.

MyYo (approved):

ThorlabsImager/yOCTHardware_closeScanner.m
```diff
@@ -1,10 +1,14 @@
-function yOCTHardware_closeScanner(v)
+function yOCTHardware_closeScanner(octSystemModule, octSystemName, skipHardware, v)
```
MyYo, line 1: Can you explain why the interface change?

MyYo (changes requested):

ThorlabsImager/yOCTHardware_closeScanner.m
alejoruge: Yes, the problem is when the hardware is accidentally initialized with the wrong system (Ganymede instead of Gan632 for example), on the second initialization call with the correct system, the code auto-detects the change and runs a teardown before re-initializing. 

That teardown called **yOCTHardware_closeScanner**, which internally called yOCTHardware('status'), but status requires **scannerInitialized = true**. 

Since the first init used the wrong system and never actually opened the scanner, scannerInitialized was false, and the teardown crashed before re-init could even start:

```
Scanner is not initialized. Call yOCTHardware('init', ...) with octProbePath.
Error in yOCTHardware_closeScanner : called status, which rejected it
Error in yOCTHardware (teardown)      : teardown called closeScanner
Error in yOCTHardware (init)                 : in it called teardown
```

So the fix I applied here is simple: **yOCTHardware_closeScanner** no longer calls status internally. It now receives the system state directly as arguments from the teardown block, which already has gOCTHardwareStatus available. 

Same pattern already applied to **yOCTHardware_initScanner** earlier.

I also added a test (**testSystemChangeTeardownWithUninitializedScanner**): inits with Ganymede (skipHardware=true, so scannerInitialized=false), then re-inits with Gan632 to reproduce the exact crash scenario and verifies it completes.

Please leet me know what you think

alejoruge:
Yes, the problem is when the hardware is accidentally initialized with the wrong system (Ganymede instead of Gan632 for example), on the second initialization call with the correct system, the code auto-detects the change and runs a teardown before re-initializing. 

That teardown called **yOCTHardware_closeScanner**, which internally called yOCTHardware('status'), but status requires **scannerInitialized = true**. 

Since the first init used the wrong system and never actually opened the scanner, scannerInitialized was false, and the teardown crashed before re-init could even start:

```
Scanner is not initialized. Call yOCTHardware('init', ...) with octProbePath.
Error in yOCTHardware_closeScanner : called status, which rejected it
Error in yOCTHardware (teardown)      : teardown called closeScanner
Error in yOCTHardware (init)                 : in it called teardown
```

So the fix I applied here is simple: **yOCTHardware_closeScanner** no longer calls status internally. It now receives the system state directly as arguments from the teardown block, which already has gOCTHardwareStatus available. 

Same pattern already applied to **yOCTHardware_initScanner** earlier.

I also added a test (**testSystemChangeTeardownWithUninitializedScanner**): inits with Ganymede (skipHardware=true, so scannerInitialized=false), then re-inits with Gan632 to reproduce the exact crash scenario and verifies it completes.

Please leet me know what you think

MyYo (approved):
Thanks looks good

PR #133, opened 2026-04-09
Add Probe File: 40x WINTER for LSM03 Scan Head

alejoruge:
**Problem**

**yOCTGetProbeIniPath** accepted two arguments (magnificationStr, otherKeyword) where otherKeyword was optional and could be either a scan head identifier (OCTP900) or nothing. Meanings were a little bit ambiguous and conflated into a single keyword. 

Additionally, magnification defaulted to '10x' when omitted, hiding missing configuration. 
Probe file names followed a format (Probe - Olympus 40x WINTER.ini) with no enforced structure.

**Solution**

Refactor yOCTGetProbeIniPath into three explicit arguments:

- **magnification** (required): '10x', '20x', or '40x': now mandatory, no silent default
- **scanHead** (required): '**OCTG**' (current Gan632 setup) or 'OCTP900' (Ganymede setup): previously implicit or mixed into otherKeyword
- variant (optional, default ''): 'WINTER', 'SUMMER', or empty: only needed when the same magnification + scanHead combination has multiple calibration files (currently 40x needs this).

The function now builds the file name dynamically from these arguments instead of a switch over concatenated strings, and validates that the resolved file actually exists, listing available probes in the error if not.

All probe .ini files are renamed to a consistent convention: 

`Probe Olympus - {magnification} - {scanHead}[ - {variant}].ini.`


MyYo (changes requested):
Is Winter + OCTP900 is different from WINTER + regular system configuration?

IF so its better to give the non OCTP900 system a name

MyYo (approved):
Looks good thanks

PR #134, opened 2026-04-10
Update yOCTImportPythonModule.m

albertchen1375:
Current Problem:

- There is no sanity check to ensure that file path is verified where matlab -> python import is protected. 

Fix:

- Added a repository integrity check in [yOCTImportPythonModule.m](vscode-file://vscode-app/c:/Users/alber/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) right after module im
port.

The bridge now reads the imported module path (module.file), normalizes it, and verifies it is under:
 repo root
or repo root/src/. 
If not, throws:
- yOCTImportPythonModule:RepositoryPathMismatch
- with actual path and expected roots in the message.


MyYo (approved):
Looks good

PR #137, opened 2026-04-24
Add safe stop/resume to yOCTProcessTiledScan

alejoruge:
**Problem**
**yOCTProcessTiledScan** had no way to recover from an interrupted run. If the process crashed or was stopped mid-way, whether during the per-frame parfor loop or during the final BigTIFF assembly, all partial work was discarded. The next run would start from scratch, repeating days of compute time on scans that were already done previously.

**Solution**
Added a resume option to **yOCTProcessTiledScan** and **yOCT2Tif**. This enables:

- **Per-frame skipping:** frames whose completion marker (y####.tif.json) already exists are skipped in both the pre-scan and inside the parfor loop.
- **Half-committed frame recovery:** a lone .tif without its .json (crash between the two MW1 writes) is deleted and reprocessed, so the final output is always consistent.
- **Deterministic clim:** the global contrast range is saved in _clim.json on first compute and reused on resume, preventing the final BigTIFF from changing between runs.
- **Atomic writes:** every artifact (BigTIFF, TifMetadata.json, _clim.json, reconstructConfig.json) is written to an _inProgress sibling and renamed to its final name only when 100% complete. A crash mid-write leaves no partially-named output to confuse a future resume.
- **Config guard:** on resume, the current inputs are compared against a saved reconstructConfig.json. Any mismatch (like a different focusSigma, dispersion or focus position) raises a clear and informative error before touching any data.
- **Already-done early exit:** if the final output exists and partialMode/ is gone, the function exits immediately with an error message rather than doing a no-op reprocess.
- **Resume summary:** prints "Already reconstructed / Newly reconstructed" frane counts to match the style of yOCTUnzipTiledScan.
- **16 unit tests cover all branches:** clean run, interrupted finalize, stale _inProgress BigTIFF, corrupted clim cache, half-committed frame, config mismatch (3 fields), and fully-done skip.

### DIFF
--- a/LoadSave/yOCT2Tif.m
+++ b/LoadSave/yOCT2Tif.m
@@ -275,6 +275,12 @@ elseif mode == 2
             continue;
         end
 
+        % If a partial .tif exists without its .json (half-committed), delete it so
+        % awsCopyFile_MW1 can create its staging subdirectory at that path
+        if in.resume && awsExist(p,'file')
+            delete(p);
+        end
+
         bits = yOCT2Tif_ConvertBitsData(data(:,:,yI),c,false);
         
         % Save Tif stack file
--- a/LoadSave/yOCT2Tif.m
+++ b/LoadSave/yOCT2Tif.m
@@ -278,7 +278,7 @@ elseif mode == 2
         % If a partial .tif exists without its .json (half-committed), delete it so
         % awsCopyFile_MW1 can create its staging subdirectory at that path
         if in.resume && awsExist(p,'file')
-            delete(p);
+            awsRmFile(p);
         end
 
         bits = yOCT2Tif_ConvertBitsData(data(:,:,yI),c,false);
--- a/LoadSave/yOCT2Tif.m
+++ b/LoadSave/yOCT2Tif.m
@@ -233,6 +233,10 @@ elseif mode == 1
             catch
                 % No staged artifacts to commit, or commit already done. Safe to ignore.
             end
+            % If MATLAB died after Mode 1 created the directory but before it finished
+            % writing inside it, a leftover directory named y####.tif/ remains.
+            % Remove those so mode 2 can re-write those frames cleanly.
+            yOCT2Tif_cleanOrphanStages(outputFilePaths{3});
         else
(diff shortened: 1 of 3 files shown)


MyYo (changes requested):
In this code you implement to resume interfaces
1. resume process tiled scans.
2. resume during reconstruction of the tiff file.

I would split those to two commits. My intuition is that process tiled scan resume is more important as it takes longer then yOCT2Tif. Is that correct?

alejoruge:
> In this code you implement to resume interfaces
> 
> 1. resume process tiled scans.
> 2. resume during reconstruction of the tiff file.
> 
> I would split those to two commits. My intuition is that process tiled scan resume is more important as it takes longer then yOCT2Tif. Is that correct?

You're right that those are the two resume surfaces, and yes, in practice the parfor stitching is typically the longer of the two (the last run took: **12 min first part** vs **8.3 min finalize**)

Meaning ~60% of the time for processing tiled scans, ~40% of the time reconstructing the tiff file as shown in these images:

Screenshot 2026-04-23 at 11 16 24 PM
Screenshot 2026-04-23 at 11 19 26 PM


I can split this into two PRs if you prefer, but I'd like to flag a concern before I do: the two phases share the partialMode/ folder and the resume flag plumbed through yOCT2Tif.

If we land the first resume PR alone, the finalize step still wipes partialMode/ at the end of every run (current behaviour), so a crash during finalize would still lose the per-frame artifacts and force a full re-stitch on the next run. 

In other words: shipping (1) without (2) gives us resume for stitching-only crashes, but any finalize-stage crash still costs us the full part of stitching work.

Maybe we can land both PRs together to merge at the same time before production use? Like just want to avoid partial deployment

In the meantime I will separate them

MyYo:
thanks
the implemented logic is complex and hard to follow. Is there something we can do to simplify?

alejoruge:
> thanks the implemented logic is complex and hard to follow. Is there something we can do to simplify?


I'll mark it as draft while l review each change and split the PR. My goal is to find room for improvement to simplify this logic. I believe a refactor would be needed if we want to achieve it properly, but will try to get options. I will mark it ready for review once done.

PR #139, opened 2026-05-07
Fix galvo phase delay at low resolution in optical path correction

alejoruge:
**PROBLEM:** 
At low pixel resolutions (20 um/pixel for exmaple), the galvo mirror's mechanical inertia causes the physical beam position to lag behind the commanded position by a fixed number of A-scan samples. 

Since **yOCTOpticalPathCorrection** evaluates its polynomial at the commanded x-position (calibrated at 1 um/pixel), this lag introduces a systematic spatial offset that scales linearly with pixel size, manifesting as a tilt in corrected B-scans (even ~**100 um** deviation at 20 um/pixel, as shown below). 

image

This makes surface detection and overview scans at low resolution very unreliable.

**SOLUTION:** 
Add a **GalvoPhaseDelay_Asamples** parameter to probe .ini files. 

**yOCTOpticalPathCorrection** now reads this value and shifts the polynomial evaluation point by N x (dx − 1 um) along the fast (x) axis, compensating for the galvo lag at any scan speed. The polynomial is still calibrated at 1 um/pixel; the shift is zero at 1 um/pixel and grows proportionally at coarser resolutions.

In this PR, only the 20x Olympus / OCTG probe has been calibrated (GalvoPhaseDelay_Asamples = 15.5). All other probes are set to 0 (no compensation) until they are individually calibrated. Setting the value to 0 is fully backwards-compatible, no behavior changes for any existing scan.

This fix is a prerequisite for reliable automated tissue surface detection and overview scans when used with the 20x OCTG probe at overview resolutions.

Some low-res scan comparisons with the fix applied and not applied:

XY visual:
image

XZ visual:
image

### DIFF
--- a/Processing/test_yOCTProcessTiledScan_createDimStructure.m
+++ b/Processing/test_yOCTProcessTiledScan_createDimStructure.m
@@ -1,0 +1,158 @@
+classdef test_yOCTProcessTiledScan_createDimStructure < matlab.unittest.TestCase
+    % Tests for the galvo phase delay X-correction in yOCTProcessTiledScan_createDimStructure.
+    %
+    %   At low pixel resolutions the galvo mirror lags behind the commanded
+    %   position, so the X coordinates written in the OCT file are wrong.
+    %   yOCTProcessTiledScan_createDimStructure fixes this by shifting
+    %   dimOneTile.x.values by N*(pixelSize_um - 1)*1e-3 mm, where N is
+    %   the GalvoPhaseDelay_Asamples value stored in the probe .ini file.
+    %   After this shift every downstream function (optical path correction,
+    %   tile stitching, surface detection) automatically receives the
+    %   correct physical beam positions.
+    %
+    %   Each test creates a simulated scan on disk, sets the desired
+    %   parameters in ScanInfo.json, calls createDimStructure, and checks
+    %   that x.values came out with the expected values.
+
+    methods(TestClassSetup)
+        function setupHardwareLib(~)
+            yOCTHardware('init', 'OCTSystem', 'Ganymede', 'skipHardware', true);
+        end
+    end
+
+    properties (Access = private)
+        TestFolder = 'tmp_galvo_phase_delay_test/'
+    end
+
+    methods (Access = private)
+        function simulateScan(testCase, nXPixels, pixelSize_um, galvoPhaseDelay_Asamples)
+            % Creates a simulated tile scan folder on disk with the given pixel
+            % size and number of X columns. If galvoPhaseDelay_Asamples is
+            % non-zero it is written into ScanInfo.json so that
+            % createDimStructure can read and apply it.
+            % Pass galvoPhaseDelay_Asamples = 0 to leave the field absent.
+
+            if exist(testCase.TestFolder, 'dir')
+                rmdir(testCase.TestFolder, 's');
+            end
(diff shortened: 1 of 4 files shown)


MyYo:
If I understand correctly, you discovered that we need to apply an x correction that depends on dx.

That means if the OCT system "thinks" a specific A-scan covers position x', it actually covers a physical position x, and the conversion is:

x = x' + galvoPhaseDelay_Asamples * (dx_um - calibrationPixelSize_um)

Is that correct?

Next, the part I don't understand is why you apply the x correction only on the z-axis and not on the x-axis. Can you clarify? I would imagine we could just fix:
inputScanDimensions.x.values = inputScanDimensions.x.values +  galvoPhaseDelay_Asamples * (dx_um - calibrationPixelSize_um) and then the rest of the calculation will be accurate.

Finally, how do you calibrate this? If you can say in a few words

MyYo (changes requested):
Asked a question above

alejoruge:
> If I understand correctly, you discovered that we need to apply an x correction that depends on dx.
> 
> That means if the OCT system "thinks" a specific A-scan covers position x', it actually covers a physical position x, and the conversion is:
> 
> x = x' + galvoPhaseDelay_Asamples * (dx_um - calibrationPixelSize_um)
> 
> Is that correct?
> 
> Next, the part I don't understand is why you apply the x correction only on the z-axis and not on the x-axis. Can you clarify? I would imagine we could just fix: inputScanDimensions.x.values = inputScanDimensions.x.values + galvoPhaseDelay_Asamples * (dx_um - calibrationPixelSize_um) and then the rest of the calculation will be accurate.
> 
> Finally, how do you calibrate this? If you can say in a few words


You're right here on both counts..

On the formula: Yes, that's exactly the relationship. The OCT system labels A-scan **i** with the commanded galvo position **x'**, but the beam was physically at x = x' − **N · (dx − 1 um)**, where 1 micron is the pixel size at which the polynomial was calibrated. 

The sign in my formula was negative (shift down) rather than positive because we're correcting the recorded position back to where the beam actually was.

On why I applied it to Z instead of X: 

Right I see the point here now more clearly. That was the first attempt and served as a a band'aid I guess. 

Shifting the polynomial evaluation point inside yOCTOpticalPathCorrection removed the visible Z tilt in B-scans, but left the X coordinates in the file unchanged. As you noted, that means tile stitching, surface detection, and TIFF metadata and all that all still operate on the wrong X values


Please see the changes here. What this updated version does: 

The correction is now applied at the single source of truth for tile X coordinates in **yOCTProcessTiledScan_createDimStructure**.m, where dimOneTile.x.values is built from ScanInfo.json. 

After the fix:

xCorrection_mm = N * (pixelSize_um - 1) * 1e-3;
dimOneTile.x.values = dimOneTile.x.values - xCorrection_mm;

Every downstream taker like yOCTOpticalPathCorrection, tile stitching in yOCTProcessTiledScan, yOCTFindTissueSurface, TIFF metadat, etc, automatically sees the physical beam position without any further changes. 

**yOCTOpticalPathCorrection** is back to its original form (no inline shift).

The correction is zero at 1 um/pixel (calibration resolution), so all existing data and all probes without a calibrated GalvoPhaseDelay_Asamples are unaffected

MyYo (approved):

PR #141, opened 2026-05-21
Add mu within the simulated scans and update test

alejoruge:
**PROBLEM**

test_yOCTEstimateScatteringCoefMuS used hand-crafted synthetic data that bypassed the OCT pipeline entirely. This meant the test was not validating the estimator in the same conditions as real usage or as the sidecar scripts (script_12_process_oct), which run through yOCTSimulateTileScan -> yOCTProcessTiledScan pipleine.

Additionally, supporting two separate tissue simulation paths (one for mu_s testing, one for everything else) would have required a new input parameter to activate the mu_s-compatible tissue model, adding complexity and a code path that only exists for testing.

SOLUTION

yOCTSimulateTileScan now always embeds a fixed synthetic tissue profile with mu_s = 10 mm-1 directly inside the simulated volume. This makes the simulator produce physically realistic exponential decay without any extra parameters or special modes.

test_yOCTEstimateScatteringCoefMuS was rewritten to use the full pipeline: yOCTSimulateTileScan -> yOCTProcessTiledScan -> yOCTFromTif, then estimate mu_s from the reconstructed volume.

Consequences of this change are intentional and documented, making updates in the test file for estimation of mu.

### DIFF
--- a/Processing/test_yOCTEstimateScatteringCoefMuS.m
+++ b/Processing/test_yOCTEstimateScatteringCoefMuS.m
@@ -27,8 +27,15 @@ classdef test_yOCTEstimateScatteringCoefMuS < matlab.unittest.TestCase
 
             if isfolder(testCase.tmpFolder), rmdir(testCase.tmpFolder, 's'); end
 
-            % Simulate OCT volume
-            dummyData = ones(zSize, xSize, ySize);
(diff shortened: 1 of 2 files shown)


alejoruge:
Updated the data sent to simulate tile scan function.

PR #143, opened 2026-05-28
Fix auto-detected scan range not being a multiple of scan tile size

alejoruge:
**PROBLEM:**
When migrating from the manual UI adjustment to the fully-automatic detection, the code that kept the returned range aligned to tile size multiples was lost by mistake.

The boundary used to clamp the detected range was changed from the user-requested overview range (e.g. [-5, 5]) to the actual reconstructed pixel extent (e.g. [-5, 4.98]). 

Since the pixel extent is not tile-aligned, the returned scan range ended up with a span like 9.98 mm instead of 10 mm, which yOCTScanTile rejected with: "the size of the scan is not a multiplier of the FOV".

**SOLUTION:**
Clamp the auto-detected range against the user-requested overview range instead of the pixel extent. The requested range is always a whole number of tiles (enforced by yOCTScanTile when the overview itself is scanned), so the returned span is always tile-aligned. This also ensures the full overview area is used when tissue spans to the edge — no coverage is lost.

### DIFF
--- a/Applications/yOCTScanTissueOverview.m
+++ b/Applications/yOCTScanTissueOverview.m
@@ -203,7 +203,7 @@ else
 end
 
 [xTissueRange_mm, yTissueRange_mm, tissueCentroid_mm, tissueArea_mm2] = detectTissueScanRange( ...
-    processedTifPath, overviewZ_um, scanTileSize_mm, temporaryFolder, v);
+    processedTifPath, overviewZ_um, scanTileSize_mm, xRange_mm, yRange_mm, temporaryFolder, v);
 
 end % main function
 
@@ -213,7 +213,7 @@ end % main function
     % tile-aligned scan range that covers it. Also saves two PNGs into
     % temporaryFolder: the raw en-face image and a summary with the accepted range.
     function [xTissueRange_mm, yTissueRange_mm, tissueCentroid_mm, tissueArea_mm2] = detectTissueScanRange( ...
-        processedTifPath, overviewZ_um, scanTileSize_mm, temporaryFolder, v)
+        processedTifPath, overviewZ_um, scanTileSize_mm, xRange_mm, yRange_mm, temporaryFolder, v)
 
     figureVisibility = ternary(v, 'on', 'off');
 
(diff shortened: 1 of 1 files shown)


Applications/yOCTScanTissueOverview.m
```diff
@@ -308,9 +308,11 @@
     if ~v, close(hRaw); end
 
     %% Compute tile-aligned scan range, clamped to the overview boundary
-    % Overview boundary comes from the loaded volume itself, not from the caller.
-    overviewXRange_mm = [min(x_mm_row), max(x_mm_row)];
-    overviewYRange_mm = [min(y_mm_col), max(y_mm_col)];
+    % Use the requested scan range as the boundary, not the actual
```
MyYo: Unclear try to rephrase it

MyYo (changes requested):
The solution in the comment is unclear

Applications/yOCTScanTissueOverview.m
alejoruge: Please see the new comment:

    % The reconstructed pixel grid ends slightly short of the requested range
    % (e.g. 4.98 instead of 5.0 mm) due to pixel discretization. Clamping to
    % that pixel extent would break tile alignment, which needs to be multiple of
    % the FOV. Use the original requested range instead, which is guaranteed 
    % to be a whole number of tiles:

Let me know what you think about it

MyYo: It's too detailed. I'm not looking for details, I just wanted to understand what you are trying to do in high level.

alejoruge: How about:

% Use the overview range as the clamping boundary as it is always
% an integer multiple of the tile FOV

or 

% Use the requested range as the clamping boundary as it is always tile-aligned.

MyYo: why there is a because?
Isn't using the clamping boundary the right thing to do period? what are the other options?

alejoruge: Right, the comment was explaining why we don't do something else instead of just saying what we do. I did it to try to explain the change (as we were using pixel grid, which is different from overview by a tiny amount and created the problem, but unnecessary for the end product)

How about this, which is just deleting the extra explanation as the current title pretty much explains it?

MyYo: The comment is great, thank you.

What I have a hard time understanding is why originally we thought:
    overviewXRange_mm = [min(x_mm_row), max(x_mm_row)];
    overviewYRange_mm = [min(y_mm_col), max(y_mm_col)];

is better then:
   overviewXRange_mm = xRange_mm;
    overviewYRange_mm = yRange_mm;

In other words, I understand what we are getting by making the switch, but it's unclear what we are loosing.
Does that make sense?

alejoruge: It felt it was more precise since those are the actual pixel center positions from the processed .tif, so they represent the actual data extent we'd use after processing. That reasoning is not wrong in the sense that the .tif pixel positions do reflect the real reconstructed volume

But it wouldn't affect the scan because the clamping boundary is only used to keep the new scan range from going outside the overview area.

Whether that boundary is [-4.99, 4.99] or [-5, 5] makes no practical difference, any tissue near the edge would be caught either way, and the 10 um difference is truly far below the resolution of the scan we're planning anyways.

So nothing is really lost, we just pick the boundary that keeps tiles aligned

MyYo: Got it so it's a simple bug that was fixed.

MyYo (approved):
Great thanks!

PR #146, opened 2026-07-09
Add B-scan averaging support for Ganymede (native DLL)

alejoruge:
**Problem**

Bscan averaging was not enabled yOCTScanTile (it just errored out). 

On Ganymede the native DLL actually does acquire and save the averaged B-scans correctly: it writes nYPixels x nBScanAvg separate Spectral{i}.data files, grouped nBScanAvg per Y position, in the order MATLAB expects. 

The only thing missing is the metadata: the DLL leaves Acquisition/SpeckleAveraging/SlowAxis = 1 in Header.xml, so MATLAB's reader believes there was no averaging and loads only the first frame of each Y position, throwing away the rest or creating wrong reconstructions afterwards.

**Solution**

Enable Bscan averaging on Ganymede by fixing that single metadata field after each scan, so MATLAB averages the repeats it already has on disk. Also expose **nBScanAvg** as a user input in the demos so it can actually be used.

- **yOCTScanTile**.m: allow Bscan averaging in the guard; after each tile is scanned, for any non-Gan632 system, call **yOCTSetBScanAveragingInHeader** to rewrite SlowAxis = nBScanAvg.

- yOCTSetBScanAveragingInHeader.m (new): rewrites the SlowAxis value in one tile's Header.xml. Idempotent and re-runnable.

- Demo_ScanAndProcess3D.m, Demo_ScanAndProcess3DOverTime.m: add an nBScanAvg input (default 1) and pass it to yOCTScanTile.

Comparison of 1 bScans vs 10 bscans in Ganymede:

Screenshot 2026-07-09 at 12 53 46 PM


ThorlabsImager/yOCTSetBScanAveragingInHeader.m
MyYo, line 1: I think this can be folded in to one of the other helper functions in ThorlabsImager

MyYo (changes requested):

ThorlabsImager/yOCTSetBScanAveragingInHeader.m
alejoruge: I moved the fix into **unzipTiledScan** as an internal helper rather than keeping it as a separate function.

The reason after analyzing the process is that some scans are intentionally not unzipped during acquisition to reduce scan time. 

In those cases, **Header.xml** is not available until the scan is later extracted, potentially on a different computer. Since the header can only be modified after extraction, handling the correction during the unzip step is the most reliable place.

**unzipTiledScan** already reads the Scan json file, so it can check whether B-scan averaging is greater than 1 and whether the system is Ganymede. When both conditions are met, it updates **Acquisition/SpeckleAveraging/SlowAxis** in the extracted Header.xml.

This way, the helper remains internal to the unzip workflow, and there is no scenario where the header correction is skipped, even when unzipping was disabled during the original scan.

alejoruge:
I moved the fix into unzipTiledScan as an internal helper rather than keeping it as a separate function.

The reason after analyzing the process is that some scans are intentionally not unzipped during acquisition to reduce scan time.

In those cases, Header.xml is not available until the scan is later extracted, potentially on a different computer. Since the header can only be modified after extraction, handling the correction during the unzip step is the most reliable place.

unzipTiledScan already reads the Scan json file, so it can check whether B-scan averaging is greater than 1 and whether the system is Ganymede. When both conditions are met, it updates Acquisition/SpeckleAveraging/SlowAxis in the extracted Header.xml.

This way, the helper remains internal to the unzip workflow, and there is no scenario where the header correction is skipped, even when unzipping was disabled during the original scan.

LoadSave/yOCTUnzipTiledScan.m
MyYo, line 1: Lets find the right home for this, it's doesn't feel like unzipping is the right spot

MyYo (changes requested):

LoadSave/yOCTUnzipTiledScan.m
alejoruge: Right! The correct home was in the native C++ DLL code. 

I now set SlowAxis = nBScanAvg in ThorlabsImagerOCT.cpp after saving the metadata and before creating the .oct file. 

I removed the workaround from the unzip workflow, rebuilt the DLL by compiling it, and tested it with nBScanAvg = 7; the extracted header now contains the correct value. Done, thank you!

alejoruge:
I think the correct home was in the native C++ DLL code. 

I now set SlowAxis = nBScanAvg in ThorlabsImagerOCT.cpp after saving the metadata and before creating the .oct file. 

I removed the workaround from the unzip workflow, rebuilt the DLL by compiling it, and tested it with nBScanAvg = 7; the extracted header now contains the correct value. Done, thank you!

Applications/yOCTScanGlassSlideToFindFocusAndDispersionQuadraticTerm.m
```diff
@@ -24,6 +25,7 @@
 addParameter(p,'focusPositionInImageZpixInitialGuess',400,@isnumeric);
 addParameter(p,'focusSearchSize_um',25,@(x)(isnumeric(x) & x>0));
 addParameter(p,'tissueRefractiveIndex',1.4);
+addParameter(p,'nBScanAvg',1,@isnumeric); % Number of B-scans to average at each position
```
MyYo: Are you sure b-scan averaging is needed for this function?
Glass slides are usually bright so no averaging is needed. Averaging can make the image less sharp

Applications/yOCTScanTissueOverview.m
```diff
@@ -18,6 +18,7 @@
 %   pixelSize_um            20          XY pixel resolution for overview (µm).
 %   scanTileSize_mm         0.5         FOV of each overview tile (mm).
 %   tissueRefractiveIndex   1.33        Tissue refractive index.
+%   nBScanAvg               1           Number of B-scans to average at each position.
```
MyYo: Same question, for scan overview I would think B Scan averaging is not needed.

Applications/yOCTTissueSurfaceAutofocus.m
```diff
@@ -19,6 +19,7 @@
 %   moveTissueToFocus: When set to true, it will move the Z stage automatically when the surface is out of focus.
 %       When set to false, it will not move stage, but will notify user how to move stage back to focus.
 %       Default is true. If skipHardware is set to true (via yOCTHardwareLibSetUp) then moveTissueToFocus is forced to false.
+%   nBScanAvg: Number of B-scans to average at each position. Default: 1.
```
MyYo: same comment as above 2 files

MyYo (approved):
Small changes. 
You can merge after addressing the comments.

Applications/yOCTScanGlassSlideToFindFocusAndDispersionQuadraticTerm.m
alejoruge: Yes, I actually agree with you. I was giving the option but now that you brought it up I don't really see a situation where it's needed. Will remove them from these scripts and merge. Thank you!

PR #148, opened 2026-07-09
Add yOCTMeasureFocusDrift to measure focus positions for RI-mismatched z-stacks

alejoruge:
**Problem**

When imaging through a refractive index (RI) mismatch, for example a water-immersion objective scanning into cleared tissue, the optical focus drifts in z-pixel space as the stage moves deeper. 

There was no tool to measure this drift. Without calibrating it, you either pass a single hardcoded focusPositionInImageZpix to yOCTProcessTiledScan (correct only at one depth) or skip focus-weighted stitching entirely, losing axial resolution throughout the stack.

**Solution**

yOCTMeasureFocusDrift allows the user to measure the focus band on a handful of reconstructed Bscans sampled across the z-stack. 

It fits a drift curve to those measurements and returns a full per=depth focusPositionInImageZpix vector: one value per z-depth in ScanInfo.json, ready to pass directly into yOCTProcessTiledScan. 

The measurement takes a few seconds and only needs to be done properly once per sample/objective combination.

Resulting zStacks with measured focus positions:

Screenshot 2026-07-09 at 12 58 44 PM


Processing/yOCTMeasureFocusDrift.m
```diff
@@ -0,0 +1,622 @@
...
+%
+% INPUTS:
+%   volumeOutputFolder          - folder containing the scans and ScanInfo.json
+%   dispersionQuadraticTerm     - dispersion correction value
+%
+% OPTIONAL PARAMETERS:
+%
+%   Parameter          Default
+%
+%   'depthSpacing_um'   50      Starting at the tissue/coverslip interface
```
MyYo: Parameter name is not clear. Can you try rephrasing?

Processing/yOCTMeasureFocusDrift.m
```diff
@@ -0,0 +1,622 @@
...
+%                               (stage z = 0), measure one tile every this
+%                               many microns of stage depth going deeper.
+%                               Shallower (negative Z, gel) scans are skipped.
+%                               Stop interactively once focus is not visible.
+%   'outputPath'        ''      Where to save results (drift figure + focus vector)
```
MyYo: outputDirectory?

MyYo (changes requested):

MyYo:
In addition, I think you need to add a tester

Processing/yOCTMeasureFocusDrift.m
alejoruge: Renamed to outputDirectory

Processing/yOCTMeasureFocusDrift.m
alejoruge: Good point, I renamed it to **focusMeasureStep_um** and reworded the description so it's clear it's the depth step at which we measure the focus: starting at the tissue surface (z=0), measure the focus every this many microns of stage depth going deeper. Please let me know what u think.

MyYo (changes requested):
Can you add a tester?

Demo_ScanAndProcess3D.m
```diff
@@ -38,6 +38,11 @@
 % Set to true if you would like to process existing scan rather than scan a new one.
 skipScanning = false;
 
+% Set to true to manually measure focus drift (a per-depth focus) before processing.)
```
MyYo, line 41: Parameter name is confusing in the context of the note.
Maybe change to autoCorrectFocusDrift?

Processing/yOCTProcessTiledScan.m
```diff
@@ -106,6 +106,35 @@ function yOCTProcessTiledScan(varargin)
 %% Load configuration file & set parameters
 json = awsReadJSON([tiledScanInputFolder 'ScanInfo.json']);
 
+focusPositionInImageZpix = in.focusPositionInImageZpix;
+% If the focus position is the same for all volumes, create a vector that
+% stores the focus position for each volume. This is mainly to enable
+% compatbility with the upgraded yOCTFindFocus which returns a focus value
+% for each volume
```
MyYo, line 113: I think you mean "for each volume"-->"for each depth"

PR #149, opened 2026-07-09
Wire focus drift measurement into Demo_ScanAndProcess3D

alejoruge:
Problem: When imaging through cleared tissue with a water objective, focus drifts with depth due to refractive index mismatch. 

Solution: Wire focus drift measurement it into Demo_ScanAndProcess3D behind a measureFocusDrift = false flag, placed after hardware teardown (dispersion already known, volume on disk) and before reconstruction.

### DIFF
--- a/Demo_ScanAndProcess3D.m
+++ b/Demo_ScanAndProcess3D.m
@@ -37,6 +37,11 @@ output_folder = '\';
 % Set to true if you would like to process existing scan rather than scan a new one.
 skipScanning = false;
 
+% Set to true to manually measure focus drift (a per-depth focus) before processing.)
+% Useful when the focus shifts with depth, e.g. a water objective imaging into
+% cleared tissue. When true it overrides focusPositionInImageZpix below:
+measureFocusDrift = false;
+
 %% Load hardware
 yOCTHardware('init', 'OCTSystem', octSystem, 'skipHardware', skipScanning, ...
     'octProbePath', octProbePath, 'v', true);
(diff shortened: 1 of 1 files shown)


Demo_ScanAndProcess3D.m
```diff
@@ -37,6 +37,11 @@
 % Set to true if you would like to process existing scan rather than scan a new one.
 skipScanning = false;
 
+% Set to true to manually measure focus drift (a per-depth focus) before processing.)
+% Useful when the focus shifts with depth, e.g. a water objective imaging into
+% cleared tissue. When true it overrides focusPositionInImageZpix below:
+measureFocusDrift = false;
```
MyYo, line 43: Is there a reason not to always measure?

MyYo (changes requested):

Demo_ScanAndProcess3D.m
alejoruge: I don’t think we should always measure it since When there is no significant refractive index mismatch, for example, when using water as the medium with uncleared tissue, there is essentially no focus drift, so a single focus value should work reliably. 

Adding multiple measurements in that case could introduce unnecessary variation and potentially lead to an incorrect reconstruction.

I would suggest measuring multiple positions only when it is actually needed, such as with transparent tissue in a medium that has a large refractive index mismatch.

alejoruge:
I don’t think we should always measure it since When there is no significant refractive index mismatch, for example, when using water as the medium with uncleared tissue, there is essentially no focus drift, so a single focus value should work reliably. 

Adding multiple measurements in that case could introduce unnecessary variation and potentially lead to an incorrect reconstruction.

I would suggest measuring multiple positions only when it is actually needed, such as with transparent tissue in a medium that has a large refractive index mismatch.

MyYo (approved):

PR #150, opened 2026-07-17
Fix Matrox non-paged memory error (55) during scan acquisitions

alejoruge:
**Problem:**

Scanning on GAN632 (Python SDK path) reliably failed with:

- **Matrox error 55:** Not enough non-paged memory to allocate the buffer.

Screenshot 2026-07-17 at 2 30 53 PM

This scan failed and was taking ~112 seconds per scan.

Acquisition used **ACQ_ORDER_ALL**, which makes the framegrabber allocate the entire oversampled volume as one Matrox buffer, and sometimes like using higher (e.g nBScanAvg=8) that's ~20 GB. This memory comes from the Windows kernel/driver non-paged pool, not regular system RAM, so having 120 GB of RAM didn't help, specially if RAM was being used for other system ops.

Worse, that buffer wasn't reliably released between tiles (a Python reference cycle between oct_file/raw_data/scan_pattern kept it alive), so the non-paged pool was exhausted after about ~5 tiles from fresh restart.

Then we manually split the scanned files to force it to look like what MATLAB uses, which takes more time than necessary.

**Solution**

By reading pyhton SDK, switched acquisition to **FRAME_BY_FRAME**, matching the C++ Ganymede implementation that never had this problem. Instead of requesting the whole volume at once, drain the acquisition one B-scan at a time and write each directly to its own Spectral{i}.data file, which is only ~4 MB is ever staged in the driver, and the post-hoc file-splitting step is no longer needed.

IMPORTANT: Removing that split also eliminates a full read-and-rewrite of the entire multi-GB volume per tile (the old step read the whole volume into memory and wrote it back out as thousands of files). 

In other words, not only it will not run into matrox errors, but speed will be 30% FASTER.

No need to use high end computing power and memory, scan overall is not affected in terms of time, and reduced process times of splitting files manually.

**Verification:**

See the resulting speed of the same exact scan that previously failed in the scan 5 with matrox error:

Screenshot 2026-07-17 at 2 48 50 PM

Same scan parameters, but now takes only **~79 seconds. Speed difference ~30% faster!**

So this full 8-tile acquisition at nBScanAvg=8 completed with no Matrox errors that previously failed consistently around tile 4–5) now with faster speed, no memory overloading, and FRAME_BY_FRAME changes data delivery order, not galvo scan speed change.

Additional hardening:

- Pre-check memory via the properly handled SDK, so an oversized request fails with a clear message instead of a cryptic Matrox error mid-acquisition explaining the memory issue (not needed with 4MB but good practice in case we ever switch it back).

- Lost-frame detection: if a frame is dropped, the tile fails cleanly instead of hanging forever waiting for data that will never arrive, and MATLAB's existing retry logic re-scans the tile if this rarely happens.

- Deterministic buffer release (gc.collect() after every tile), since the reference cycle prevented normal refcounting from freeing the SDK's native buffers in time.

### DIFF
--- a/ThorlabsImager/ThorlabsImagerPython/thorlabs_imager_oct.py
+++ b/ThorlabsImager/ThorlabsImagerPython/thorlabs_imager_oct.py
@@ -209,13 +209,13 @@ def yOCTScan3DVolume(centerX_mm: float, centerY_mm: float,
     scan_pattern = None
     raw_data = None
     oct_file = None
+    frames = None
     acquisition_started = False
     
     try:
         # Set B-scan averaging on probe and processing
-        if nBScanAvg > 1:
-            _probe.properties.set_oversampling_slow_axis(nBScanAvg)
-            _processing.properties.set_bscan_avg(nBScanAvg)
+        _probe.properties.set_oversampling_slow_axis(nBScanAvg)
+        _processing.properties.set_bscan_avg(nBScanAvg)
         
         # Create volume scan pattern
         scan_pattern = _probe.scan_pattern.create_volume_pattern(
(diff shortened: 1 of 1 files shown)


MyYo (approved):
Good finding. Excellent 30% improvement

PR #153, opened 2026-08-20
Protect existing outputs from accidental overwrite in yOCTProcessTiledScan

alejoruge:
**Problem:**

yOCTProcessTiledScan deletes the existing output file as soon as processing starts (yOCT2Tif partial file mode init), and the new file only materializes at the end. 

As some people like Komal yesterday have accidentally re-run a script that processes to the same path (like Demo_ScanAndProcess3D) by mistake, the previous reconstruction is destroyed within seconds, including reconstructions that took many hours to build times with no way to recover it.

**Solution**

Before processing starts, if an output path (file or folder) already exists, rename the old one to <name>_old_<date> and print a notice. The new result is still saved to the requested path.

### DIFF
--- a/Processing/yOCTProcessTiledScan.m
+++ b/Processing/yOCTProcessTiledScan.m
@@ -93,6 +93,25 @@ if ischar(outputPath)
     outputPath = {outputPath};
 end
 
+% Protect existing outputs from accidental overwrite. If an output already
+% exists, rename the old one to <name>_2 (or _3, ...). The new result is
+% still saved to the requested path (local paths only).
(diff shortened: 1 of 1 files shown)


Processing/yOCTProcessTiledScan.m
MyYo, line 1: Move to the script, not in the function

MyYo (changes requested):

Processing/yOCTProcessTiledScan.m
alejoruge: Got it. Resolved.

